PR #1, opened 2021-11-04
Create rust.yml

dburgener:


.github/workflows/rust.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup
+      run: |
+              sudo apt-get update -q
+              sudo apt-get -y install secilc
+    - name: Build
+      run: cargo build --verbose
```
l0kod, line 24: We should reuse [build.sh](https://github.com/dburgener/cascade/blob/main/.pipelines/build.sh) and similar scripts if possible, or at least their content (e.g. `rustup run stable cargo build --verbose`).

.github/workflows/rust.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Run tests
+      run: cargo test --verbose
+    - name: Check format
+      run: cargo fmt --all -- --check
```
l0kod, line 28: Other steps could be inspired from https://github.com/dburgener/cascade/blob/main/.pipelines/OneBranch.PullRequest.yml#L48-L85

l0kod (changes requested):

l0kod:
As suggested by @liuw (see (comment)[https://github.com/dburgener/cascade/pull/2#issuecomment-961509579]) we should also use `cargo clippy`. It would require to exclude the generated `parser.rs` from the analysis though.

dburgener:
> As suggested by @liuw (see comment) we should also use cargo clippy. It would require to exclude the generated parser.rs from the analysis though, see lalrpop/lalrpop#632 (which is not yet published in crates.io).

I definitely agree with this. We should do it in a follow-up PR though so that we can get at least some CI working ASAP.

l0kod (approved):

PR #2, opened 2021-11-04
Fix clippy warnings

liuw:
This PR fixes a lot of clippy warnings. Code is now much more idiomatic.

No functional change intended. The code still builds and all tests pass after these changes.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -289,13 +289,13 @@ impl FuncCall {
         if let Some(_) = self.class_name {
             return None;
         }
-        if constants::AV_RULES.iter().any(|i| *i == &self.name) {
+        if constants::AV_RULES.iter().any(|i| *i == self.name) {
             return Some(BuiltIns::AvRule);
         }
-        if &self.name == constants::FILE_CONTEXT_FUNCTION_NAME {
+        if self.name == constants::FILE_CONTEXT_FUNCTION_NAME {
             return Some(BuiltIns::FileContext);
         }
-        if &self.name == constants::DOMTRANS_FUNCTION_NAME {
+        if self.name == constants::DOMTRANS_FUNCTION_NAME {
             return Some(BuiltIns::DomainTransition);
         }
         None
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -286,7 +286,7 @@ impl FuncCall {
     }
 
     pub fn check_builtin(&self) -> Option<BuiltIns> {
-        if let Some(_) = self.class_name {
+        if self.class_name.is_some() {
             return None;
         }
         if constants::AV_RULES.iter().any(|i| *i == self.name) {
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -23,7 +23,7 @@ impl fmt::Display for HLLString {
 impl HLLString {
     pub fn new(string: String, range: Range<usize>) -> Self {
         HLLString {
-            string: string,
+            string,
             range: Some(range),
         }
     }
@@ -128,10 +128,7 @@ pub struct PolicyFile {
 
 impl PolicyFile {
     pub fn new(policy: Policy, file: SimpleFile<String, String>) -> Self {
-        PolicyFile {
-            policy: policy,
-            file: file,
-        }
+        PolicyFile { policy, file }
     }
 }
 
@@ -142,7 +139,7 @@ pub struct Policy {
 
 impl Policy {
     pub fn new(exprs: Vec<Expression>) -> Policy {
-        Policy { exprs: exprs }
+        Policy { exprs }
     }
 }
 
@@ -208,8 +205,8 @@ pub struct TypeDecl {
 impl TypeDecl {
     pub fn new(name: HLLString, inherits: Vec<HLLString>, exprs: Vec<Expression>) -> TypeDecl {
         TypeDecl {
-            name: name,
-            inherits: inherits,
+            name,
+            inherits,
             is_virtual: false,
             expressions: exprs,
             annotations: Annotations::new(),
@@ -344,7 +341,7 @@ pub struct Annotation {
 impl Annotation {
     pub fn new(name: HLLString) -> Self {
         Annotation {
-            name: name,
+            name,
             arguments: Vec::new(),
         }
     }
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -151,9 +151,8 @@ pub enum Expression {
 
 impl Expression {
     pub fn set_class_name_if_decl(&mut self, name: HLLString) {
-        match self {
-            Expression::Decl(Declaration::Func(d)) => d.class_name = Some(name),
-            _ => (),
+        if let Expression::Decl(Declaration::Func(d)) = self {
+            d.class_name = Some(name)
         }
     }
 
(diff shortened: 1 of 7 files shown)


liuw:
Feel free to do anything to these patches, including but not limited to rejecting all of them. I know it is quite a bit of code churn, but I just couldn't resist. :-)

If you end up taking these patches, I think it would also be good to setup `cargo clippy` check in the CI.

src/lib.rs
```diff
@@ -194,10 +194,12 @@ mod tests {
                     policy_path
                         .with_extension("cil")
                         .file_name()
-                        .expect(&format!(
-                            "failed to extract file name from `{}`",
-                            policy_path.to_string_lossy()
-                        )),
+                        .unwrap_or_else(|| {
```
l0kod, line 197: I don't get this suggestion, which version of Clippy do you use?

src/lib.rs
```diff
@@ -290,7 +292,7 @@ mod tests {
         for name in &["a", "a_a", "a_a_a", "a_aa_a", "a0", "a_0", "a0_00"] {
             let _: ast::HLLString = parser::NameDeclParser::new()
                 .parse(name)
-                .expect(&format!("failed to validate `{}`", name));
+                .unwrap_or_else(|_| panic!("failed to validate `{}`", name));
```
l0kod, line 295: A bit more verbose but if Clippy said so…

src/lib.rs
liuw: clippy 0.1.56 (59eed8a 2021-11-01)

l0kod:
Thanks a lot for these clean commits @liuw! It looks good to me.

It seems that there is some remaining useless return statements which are not addressed (https://github.com/dburgener/cascade/blob/main/src/internal_rep.rs#L622), not sure if you forgot it or if we have different version of Clippy.

liuw:
> Thanks a lot for these clean commits @liuw! It looks good to me.
> 
> It seems that there is some remaining useless return statements which are not addressed (https://github.com/dburgener/cascade/blob/main/src/internal_rep.rs#L622), not sure if you forgot it or if we have different version of Clippy.

I did see that, but I somehow forgot to change that site. I will add an additional commit to this series.

liuw:
BTW there are two things that I have not addressed.

One looks severe, but I did not understand the code well enough to make a judgement call.

```
error: you are deriving `Hash` but have implemented `PartialEq` explicitly
   --> src/ast.rs:195:28
    |
195 | #[derive(Clone, Debug, Eq, Hash)]
    |                            ^^^^
    |
note: `PartialEq` implemented here
   --> src/ast.rs:217:1
    |
217 | / impl PartialEq for TypeDecl {
218 | |     fn eq(&self, other: &Self) -> bool {
219 | |         self.name == other.name
220 | |     }
221 | | }
    | |_^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#de
rive_hash_xor_eq
    = note: this error originates in the derive macro `Hash` (in Nightly builds, run with -Z macro-bac
ktrace for more info)
```

The other one is about adding default implementation for types, which I think is a bit subjective at this point.

One example:

```
warning: you should consider adding a `Default` implementation for `Annotations`
   --> src/ast.rs:363:5
    |
363 | /     pub fn new() -> Self {
364 | |         Annotations {
365 | |             annotations: Vec::new(),
366 | |         }
367 | |     }
    | |_____^
    |
    = note: `#[warn(clippy::new_without_default)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ne
w_without_default
help: try adding this
    |
358 + impl Default for Annotations {
359 +     fn default() -> Self {
360 +         Self::new()
361 +     }
362 + }
```

src/lib.rs
l0kod: Indeed, the --fix command includes this but there is no trace in the printed warnings… The nightly Clippy (0.1.58) adds some more fixes if you want to take them.

liuw: Yep. I did most of the fixes by hand and carefully arranged them into groups because I wanted to make this series reviewable by humans.

I got lazy near the end seeing there was only one file remaining, so I did `cargo clippy --fix` instead for `lib.rs`.

If I have more time today, I may be able to have look at warnings from Clippy 0.1.58.

l0kod:
> BTW there are two things that I have not addressed.
> 
> One looks severe, but I did not understand the code well enough to make a judgement call.
> 
> ```
> error: you are deriving `Hash` but have implemented `PartialEq` explicitly
>    --> src/ast.rs:195:28
>     |
> 195 | #[derive(Clone, Debug, Eq, Hash)]
>     |                            ^^^^
>     |
> note: `PartialEq` implemented here
>    --> src/ast.rs:217:1
>     |
> 217 | / impl PartialEq for TypeDecl {
> 218 | |     fn eq(&self, other: &Self) -> bool {
> 219 | |         self.name == other.name
> 220 | |     }
> 221 | | }
>     | |_^
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#de
> rive_hash_xor_eq
>     = note: this error originates in the derive macro `Hash` (in Nightly builds, run with -Z macro-bac
> ktrace for more info)
> ```

This should not an issue right now but I agree that it can lead to inconsistencies and we should fix it. The simpler fix would be to only use `BTreeSet` instead of `HashSet` for the synthetic expressions (see `global_exprs` and `local_exprs` in compile.rs).

> 
> The other one is about adding default implementation for types, which I think is a bit subjective at this point.

Right, that makes sense but it could be more verbose.

src/ast.rs
```diff
@@ -287,9 +286,8 @@ impl FuncCall {
     }
 
     pub fn check_builtin(&self) -> Option<BuiltIns> {
-        match self.class_name {
-            Some(_) => return None,
-            None => (),
+        if let Some(_) = self.class_name {
```
dburgener: This could even be written more tersely as `if self.class_name.is_some() {` I think.

liuw: Fixed in a later commit.

dburgener: Oh, nevermind.  I'm reviewing by commit and see now that that change came in later.

dburgener:
This looks great, thanks!

The only thing I'm skeptical about is the expect to unwrap_or_else change.  What's clippy's justification for that?

liuw:
> This looks great, thanks!
> 
> The only thing I'm skeptical about is the expect to unwrap_or_else change. What's clippy's justification for that?

I got lazy for `lib.rs`. That hunk was produced by `cargo clippy --fix`. Like @l0kod said, clippy did not provide a justification for it.

dburgener:
I think my preference would be to take everything except those two expect changes.  The version clippy proposes feels worse to me.  When we add clippy to the CI, we can make a more careful decision, but it seems sensible to me to take all the obviously good stuff now and then deal with the details later.

Would you mind updating that patch to drop the two expect changes and then I'll take the whole series?

liuw:
> I think my preference would be to take everything except those two expect changes. The version clippy proposes feels worse to me. When we add clippy to the CI, we can make a more careful decision, but it seems sensible to me to take all the obviously good stuff now and then deal with the details later.
> 
> Would you mind updating that patch to drop the two expect changes and then I'll take the whole series?

Sure. I've dropped those two hunks.

dburgener (approved):

dburgener:
Wonderful!  Thanks for the contribution!

PR #3, opened 2021-11-05
Fix more clippy warnings

liuw:


src/sexp_internal.rs
```diff
@@ -13,10 +13,7 @@ pub fn display_cil(expr: &sexp::Sexp) -> String {
         Sexp::List(l) => {
             format!(
                 "({})",
-                l.iter()
-                    .map(display_cil)
-                    .collect::<Vec<String>>()
-                    .join(" ")
+                l.iter().map(display_cil).collect::<Vec<String>>().join(" ")
```
l0kod, line 16: Could you please squash this into the previous commit that changes the map call?

src/error.rs
```diff
@@ -198,6 +198,12 @@ pub struct HLLErrors {
     errors: Vec<HLLErrorItem>,
 }
 
+impl Default for HLLErrors {
```
l0kod, line 201: It seems weird to implement unused code just to please Clippy.

liuw: I'm fine with dropping this patch. When clippy is enabled in the CI you will need to explicitly disable that lint.

dburgener (approved):

l0kod:
Thanks!

PR #7, opened 2022-01-14
Add symbol binding

dburgener:
This PR adds support for global symbol binding using the "let" keyword.


src/internal_rep.rs
```diff
@@ -644,13 +742,34 @@ impl<'a> ClassList<'a> {
     }
 
     pub fn is_perm(&self, perm: &str) -> bool {
+        if self.perm_sets.get(perm).is_some() {
```
l0kod, line 745: Could you add tests for this?

l0kod (commented):
Nice feature!

In the second commit description, there is a typo in "statment".

data/policies/let.cas
```diff
@@ -0,0 +1,7 @@
+let read_file_perms = [ read open getattr ];
+
+resource bar {}
+
+domain foo {
+	allow(foo, bar, file, read_file_perms);
```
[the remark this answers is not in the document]
dburgener: I've added non-list and annotations (will push later today).  I'm not sure what you mean by "multiple references".  I'm probably just missing something obvious?

src/internal_rep.rs
```diff
@@ -80,6 +98,25 @@ impl TypeInfo {
...
+        name: HLLString,
+        variant: &str,
+        file: &SimpleFile<String, String>,
+        bound_type: BoundTypeInfo,
+    ) -> Result<TypeInfo, HLLErrors> {
+        Ok(TypeInfo {
+            name: name,
+            inherits: vec![variant.into()], // Does this need to somehow grab the bound parents? Does this work for the single case?
+            is_virtual: true,               // Maybe?
+            list_coercion: false,           // TODO
```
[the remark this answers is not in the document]
dburgener: Cleaned up all the TODOs here except the decl, thanks.  I *think* None should be an acceptable value for Decl at least for now, although maybe in the long term we want to switch this into an enum instead of an option so that decl can be other things that a TypeDecl (for example a LetBinding).  I don't really see a reason this needs to be a priority right now though.

src/internal_rep.rs
```diff
@@ -158,6 +215,39 @@ impl From<&TypeInfo> for Option<sexp::Sexp> {
...
+        None => todo!(), // TODO: Return error
+    };
+
+    for ti in type_slice {
+        let ti_variant = match ti.get_built_in_variant(types) {
+            Some(v) => v,
+            None => return Err(HLLErrorItem::Internal(HLLInternalError {}).into()),
+        };
+        if ti_variant != first_type_variant {
+            unimplemented!() // TODO: Return error
```
l0kod, line 242: Probably a `todo!()`.

src/internal_rep.rs
```diff
@@ -644,13 +745,34 @@ impl<'a> ClassList<'a> {
...
+    pub fn insert_perm_set(&mut self, set_name: &str, perms: Vec<String>) {
+        self.perm_sets.insert(set_name.to_string(), perms);
+    }
+
+    pub fn expand_perm_list(&self, perms: Vec<&HLLString>) -> Vec<HLLString> {
+        let mut ret = Vec::new();
+        for p in perms {
+            if let Some(pset) = self.perm_sets.get(&p.to_string()) {
+                let pset_strings: Vec<HLLString> =
+                    pset.iter().map(|s| HLLString::from(s.as_str())).collect();
```
l0kod, line 768: Not sure `.as_str()` is required.

dburgener: There may be other solutions, but just removing the as_str() call gives:

 the trait `From<&String>` is not implemented for `HLLString`

src/compile.rs
```diff
@@ -157,6 +161,59 @@ pub fn get_built_in_types_map() -> TypeMap {
...
+                    let ti_vec = argument_to_typeinfo_vec(&v, types, classlist, None, file)?;
+                    let variant = type_slice_to_variant(&ti_vec, types)?;
+                    (
+                        variant.name.as_ref(),
+                        BoundTypeInfo::List(v.iter().map(|s| s.to_string()).collect()),
+                    )
+                }
+                a => {
+                    let ti = argument_to_typeinfo(&a, types, classlist, None, file)?;
+                    if ti.name.as_ref() == "perm" {
```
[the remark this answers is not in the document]
dburgener: I actually kind of prefer something more global here, since "perm" has special meaning in the code but not the Cascade grammar.  I think I'd prefer to do that as a follow-on PR though since it could probably use some independent discussion on it's own.

src/internal_rep.rs
l0kod: Right. Another way to do it would be to have another `From<HLLString>` impl for this, but it's OK for this PR.

PR #8, opened 2022-01-21
Add Clippy to CI

dburgener:
Fix one clippy issue and disable new_without_default


l0kod (approved):
Good! This enforces what was discussed in #2.

PR #9, opened 2022-04-13
Add APIs

dburgener:
An API is a block of functions not specifically tied to a particular type


dburgener:
Just browsing the CI here, I notice that there is a clippy warning:

```
 warning: unnecessary use of `to_string`
   --> src/compile.rs:301:35
    |
301 |         if visited_types.contains(&p.to_string() as &str) || *p == type_to_check.name {
    |                                   ^^^^^^^^^^^^^^ help: use: `p.as_ref()`
    |
    = note: `#[warn(clippy::unnecessary_to_owned)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_to_owned

warning: `selinux-cascade` (lib) generated 1 warning
```

This is interesting to me for two reasons:

1. I don't get it locally
2. The warning didn't fail the CI

Both of those probably warrant further investigation.

dburgener:
The answer to number 1 seems to be that this particular lint was added in a later version of clippy than I'm running locally.

The second issue seems likely to be that warnings are not treated as errors by default.  Per the man page, that behavior can be enabled via "--deny warnings".  I'll submit a PR to fix this warning and treat clippy warnings as errors.

dburgener:
See #11 for fix on above mentioned clippy issue.

l0kod:
> The answer to number 1 seems to be that this particular lint was added in a later version of clippy than I'm running locally.

We pinned clippy to stable, but stable is still moving. I guess that is better than forgetting to update clippy to a fixed version.

dburgener:
> We pinned clippy to stable, but stable is still moving. I guess that is better than forgetting to update clippy to a fixed version.

Yeah, I think this is all reasonable, and I updated clippy locally, so I'm back in sync now.  It was just surprising to me is all, but I agree that automatically keeping clippy up to date is worth any short term pain of new lints breaking unrelated PRs.  It's not too high effort to throw together a quick PR to either fix the lints or disable them if/when that happens.

l0kod:
Why naming this kind of helper "api"? This looks like a function or at least a (clean) macro to me.

dburgener:
> Why naming this kind of helper "api"? This looks like a function or at least a (clean) macro to me.

Well, it's a wrapper around functions.

There are two purposes here, although clarifying is drawing my intention to the fact that only the first is implemented:

1. Grouping a collection of related functions for later grouping into a policy module or system build (Cascade will support multiple different build targets out of the same repo, and these related functions can be included or not based on definition)
2. Inheriting in a domain or resource.  In this situation the API functions basically identically to a domain or resource, but with more limited functionality.  It doesn't support resource association, and it doesn't result in any actual types declared, which means that members of the api can't be referenced anywhere.  (As opposed to virtual types where the children *can* be referenced in policy.

Fundamentally, I think it's a highly similar feature to rust traits.

Point 2 is not implemented in this PR, but probably should be, so I may want to take this back and add that functionality prior to merge.

It's also potentially a valid point to argue that this should just be a special case of virtual types rather than it's own separate block.

dburgener:
Closing to reanalyze/discuss based on Mickael's concerns

PR #10, opened 2022-04-13
Support named args passing with "name=foo" in parser

dburgener:
Nothing is actually done with these yet.  Future work will need to
override the default ordering assumptions.

In our existing docs, we have used this syntax extensively in
annotations, and (to my knowledge) never in function calls.  For
consistency, we should support both.

The near term need is to support this syntax in annotations.  A
follow-on PR will add actual support here.  Allowing these to work in
the parser allows annotations that have no behavior defined yet, but use
the "name=foo" syntax to at least parse correctly.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -415,6 +415,7 @@ impl Annotations {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Argument {
     Var(HLLString),
+    Named(HLLString, Box<Argument>),
     List(Vec<HLLString>),
     Quote(HLLString),
 }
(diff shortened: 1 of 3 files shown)


l0kod:
Could you please include an example?

dburgener:
> Could you please include an example?

Can you clarify this ask?  The reason I'm asking is because this is only parser support, not full end to end support, so a unit test doesn't make a lot of sense.

If you're looking for documentation, there should be some in the existing docs, and I can look to point you there (or add it if that's not the case).

If you're looking for an example in the github comments or commit description, I can do that.  Just not sure what you're lookign for.

l0kod (approved):

l0kod:
I misunderstood the the goal. It looks good.

PR #11, opened 2022-04-13
Fail CI on clippy warnings

dburgener:


l0kod:
Could you move commit https://github.com/dburgener/cascade/pull/11/commits/df97f6c8a1f229c864e66a093cf03912330fe2e2 into #9?

l0kod:
Well, actually squashing the commit into one of #9 would be better.

dburgener:
> Could you move commit https://github.com/dburgener/cascade/commit/df97f6c8a1f229c864e66a093cf03912330fe2e2 into https://github.com/dburgener/cascade/pull/9?

I'm confused.  Why would that be preferred?  This is broken totally independently of #9. I just happened to notice it when posting #9.

l0kod:
That was to avoid warning on PR updates and make sure that, from now on, all PR will make clippy happy.

PR #12, opened 2022-04-13
Add error recovery to parser

dburgener:
When we encounter an error, we skip until the start of the next
annotated expression and continue.  If any errors were encountered, we
report and fail the build.

This enables displaying all parse errors at once rather than one at a
time.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -154,6 +154,10 @@ impl Policy {
 pub enum Expression {
     Decl(Declaration),
     Stmt(Statement),
+    // Needed for parser grammar.  If this is ever set, we should have bailed
+    // prior to using this. If possible encountering this should be an internal
+    // error, otherwise it is safe to ignore.
+    Error,
 }
 
 impl Expression {
@@ -167,6 +171,7 @@ impl Expression {
         match self {
             Expression::Decl(d) => d.add_annotation(annotation),
             Expression::Stmt(s) => s.add_annotation(annotation),
+            Expression::Error => (),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,9 +48,11 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::HL
         };
         let p = match parse_policy(&policy_str) {
             Ok(p) => p,
-            Err(e) => {
-                // TODO: avoid String duplication
-                errors.add_error(error::HLLParseError::new(e, f.into(), policy_str.clone()));
+            Err(evec) => {
+                for e in evec {
+                    // TODO: avoid String duplication
+                    errors.add_error(error::HLLParseError::new(e, f.into(), policy_str.clone()));
+                }
                 continue;
             }
         };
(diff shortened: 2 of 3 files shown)


src/ast.rs
```diff
@@ -154,6 +154,10 @@ impl Policy {
 pub enum Expression {
     Decl(Declaration),
     Stmt(Statement),
+    // Needed for parse grammar.  If this is ever set, we should have bailed
```
l0kod, line 157: "for parsed" or "to parse"?

src/ast.rs
```diff
@@ -154,6 +154,10 @@ impl Policy {
...
+    // prior to using this. If possible encountering this should be an internal
+    // error, otherwise it is safe to ignore
```
l0kod, line 159: nit: final dot.

src/lib.rs
```diff
@@ -164,10 +166,31 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::HL
...
-) -> Result<Box<Policy>, ParseError<usize, lalrpop_util::lexer::Token<'a>, &'static str>> {
+) -> Result<Box<Policy>, Vec<ParseError<usize, lalrpop_util::lexer::Token<'a>, &'static str>>> {
+    let mut errors = Vec::new();
     // TODO: Probably should only construct once
     // Why though?
-    parser::PolicyParser::new().parse(policy)
+    let parse_res = parser::PolicyParser::new().parse(&mut errors, policy);
+    // errors is a vec of ErrorRecovery.  ErrorRecovery is a struct wrapping a ParseError
+    // and a sequence of discarded characters.  We don't need those characters, so we just
+    // remove the wrapping
```
l0kod, line 176: nit: final dot.

src/lib.rs
```diff
@@ -164,10 +166,31 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::HL
...
+    let mut parse_errors: Vec<ParseError<usize, lalrpop_util::lexer::Token, &str>> =
+        errors.iter().map(|e| e.error.clone()).collect();
+    match parse_res {
+        Ok(p) => {
+            if !errors.is_empty() {
+                parse_errors.reverse();
```
l0kod, line 182: Why do we need to reverse the order here?

src/lib.rs
```diff
@@ -287,39 +310,46 @@ mod tests {
 
     #[test]
     fn basic_expression_parse_test() {
-        let res = parser::ExprParser::new().parse("domain foo {}");
+        let mut errors = Vec::new();
+        let res = parser::ExprParser::new().parse(&mut errors, "domain foo {}");
```
l0kod, line 314: Not sure about that but it may be cleaner no move the `errors` argument from `parse()` to `new()`. It depends if `ExprParser` is usually stored in a variable and if multiple methods similar to `parse()` could be added. Making `ExprParser` owns `errors` could be an issue though.

src/ast.rs
dburgener: Intention was "parse grammer" as a noun phrase.  I'll update to "parser grammar" which may be more accurate.

src/lib.rs
dburgener: I don't *think* lalrpop exposes a way to do that.  This seems to be the recommended approach in their docs [here](http://lalrpop.github.io/lalrpop/tutorial/008_error_recovery.html) and [here](http://lalrpop.github.io/lalrpop/tutorial/009_state_parameter.html).

l0kod (approved):

PR #13, opened 2022-04-14
Update to clap version 3

dburgener:
Switch to new Derive API with more intuitive syntax.

This updates the -h output slightly.  The content is the same, but the
formatting is slightly different and reordered.  It also displays
colored output now.

### DIFF
--- a/src/bin/casc.rs
+++ b/src/bin/casc.rs
@@ -3,23 +3,20 @@
 use selinux_cascade::compile_system_policy;
 use selinux_cascade::error::HLLErrorItem;
 
-use clap::{crate_authors, crate_version, App, Arg};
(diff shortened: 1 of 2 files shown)


src/bin/casc.rs
```diff
@@ -3,11 +3,19 @@
...
 
+#[derive(Parser, Debug)]
+#[clap(author, version, name = "casc")]
+struct Args {
+    #[clap(required(true))]
+    input_file: Vec<String>,
+}
+
 fn main() -> std::io::Result<()> {
+    /*
```
l0kod, line 18: Why this commented code?

l0kod (approved):

PR #16, opened 2022-04-22
Dburgener/alias

dburgener:


src/internal_rep.rs
```diff
@@ -23,7 +23,36 @@ const DEFAULT_OBJECT_ROLE: &str = "object_r";
...
+
+    pub fn insert(&mut self, key: String, value: TypeInfo) {
+        self.types.insert(key, value);
+    }
+
+    pub fn values(&self) -> std::collections::btree_map::Values<'_, String, TypeInfo> {
+        self.types.values()
+    }
+
+    pub fn iter(&self) -> std::collections::btree_map::Iter<'_, String, TypeInfo> {
```
l0kod: It would be more idiomatic to implement an iterator for `TypeMap`.

src/internal_rep.rs
```diff
@@ -367,12 +368,29 @@ fn get_type_annotations(
...
+                    }
+                }
+            }
             _ => {
                 return Err(HLLCompileError::new(
                     "Unknown annotation",
                     file,
                     annotation.name.get_range(),
-                    "The only known annotations are '@makelist' and '@associate'.",
+                    "This is not a valid annotation name.",
```
l0kod, line 439: I guess we could extend `AnnotationInfo` to get a string of the handled annotations that could help here.

data/error_policies/alias.cas
```diff
@@ -0,0 +1,5 @@
+@alias("foo")
+resource bar  {}
```
l0kod, line 2: nit: Two spaces.

l0kod:
What is the use case for this alias feature?

dburgener:
> What is the use case for this alias feature?

There are a few different use cases here.  This duplicates a feature in the existing policy language.  The main things the existing feature is used for is interoperability.

For example in a backwards compatibility scenario, you may want to rename a type, but install the new policy on a system where there may be existing objects labeled with that type.  One way to make for a clean transition is to change the name, but have an alias to the old name.  Then you can deploy the new policy with both names valid during the transition period, then update the labeling, then finally remove the alias once the old labels are gone.

Another use case is for providing an alternate scheme for naming when the kernel name for a permission or concept doesn't match the user-side name.  For example, the LSM hooks have the `setattr` permission for changing attributes on a file-like object.  From a user perspective that's often a `chown` or `chmod`.  Similarly, the `getattr` permission is equivalent to the permission needed for `stat()`.  If you create a function allowing such a permission, particularly in a broadly consumable high level abstraction, it's impossible to please everyone with a single name.  So having multiple ways to refer to it can potentially broaden the userbase.  (This second example isn't possible in the existing policy language).

src/internal_rep.rs
dburgener: Perhaps.  I worry that the message will get wordy as we keep adding annotations.  We currently have plans and/or documentation for 6 annotations (@makelist, @associate, @alias, @derive, @hint, @ignore).  Even just those six makes for a wordy message, and I would be shocked if that list doesn't grow more in the future.

src/internal_rep.rs
dburgener: This is a good idea.  I've implemented something along these lines and pushed it just now.  I tried to actually define the types inside the AliasMap impl block, and then name them just "Iter" etc, but I am getting compiler errors around this message: https://github.com/rust-lang/rust/issues/44265

It sounds like the use of generics in defining a custom type may not be fully supported, but I'm not sure as there is a lot of background reading here.

Any suggestions to make this more idiomatic from what I've just updated to?

dburgener:
TODO for myself:  The above info needs to be documented in the repo

src/internal_rep.rs
```diff
@@ -1387,6 +1422,27 @@ impl<'a> FunctionInfo<'a> {
         self.body = Some(new_body);
         errors.into_result(())
     }
+
+    // Generate the sexp for a synthetic alias function calling the real function
+    pub fn generate_synthetic_alias_call(&self, alias_cil_name: &str) -> sexp::Sexp {
```
l0kod, line 1431: Do we really want a `call baz-read` or a *copy* of baz-read, something like the synthetic resources (associations), which would be generic?

dburgener:
> I think you should be able to write a function returning impl Iterator<XXX> instead of std::collections::btree_set::Iter<XXX>: https://doc.rust-lang.org/std/collections/btree_map/struct.Iter.html#impl-Iterator
> However, that could make the type inference too generic and then require more verbose code to use this kind of iterator.

This sounds like it's still referring to the old version of the code?  I think I've done essentially that in the latest update?

(For some reason github isn't letting me respond directly to that comment)

src/internal_rep.rs
dburgener: I'm not sure I see the benefit in copying `baz-read` here.  Calling it reduces the size of the outputted policy.  The original function is guaranteed to exist because there's no way to create an alias without the "real" function.  Is there some benefit of making a new copy rather than just calling the original that I'm missing?

l0kod (approved):

PR #17, opened 2022-04-22
Add support for using the virtual keyword on functions

dburgener:
The virtual keyword indicates that a function should not be called on
that type.  It should instead be inherited and called on children.


l0kod (commented):
What should be the behavior with inherited virtual resource of method? Could you add a test with this cases?

dburgener:
> What should be the behavior with inherited virtual resource of method? Could you add a test with this cases?

I don't understand the question.  Can you clarify?  Possibly with a policy example?

dburgener:
> What about resource association? This should be covered by tests too

I think I've addressed all your comments except this one.  Trying to think through what scenarios would be appropriate here.  I'll look to add a test tomorrow if I have time.

dburgener:
Okay, one very deep dive into associated resources later, I think I've cleaned things up on the virtual functions/associated resources front.

There was a related bug where synthetic resources should become concrete when created for child concrete types that I fixed as well.

While working on this, I noticed that concrete functions for non-synthetic types don't inherit yet.  I'll plan on adding that as part of the derive work.

data/error_policies/virtual_function_non_define.cas
```diff
@@ -0,0 +1,5 @@
+virtual resource foo_parent {
+	virtual fn foo_func(domain source) {}
+}
+
+virtual resource foo inherits foo_parent {} // Illegal non-inheritance
```
l0kod, line 5: I guess people may want to automatically inherit a virtual `foo_func()`  in `foo` but I think indeed that making them explicit is cleaner, similar to Rust traits.

src/parser.lalrpop
```diff
@@ -74,7 +74,7 @@ BuiltInType: HLLString = {
 }
 
 FuncDecl: Box<FuncDecl> = {
-	"fn" <n: NameDecl> "(" <a: Comma<FuncDeclArg>> ")" "{" <b: Stmt+> "}" => Box::new(FuncDecl { class_name: None, name: n, args: a, body: b, annotations: Annotations::new() }),
+	"fn" <n: NameDecl> "(" <a: Comma<FuncDeclArg>> ")" "{" <b: Stmt*> "}" => Box::new(FuncDecl::new(n, a, b)),
```
l0kod, line 77: We could allow or force a final `;` instead of an empty `{ }`. I think it would be cleaner to deny an empty body, similarly to a function signature declaration in C or Rust.

l0kod (commented):
Nice improvement!

data/error_policies/virtual_function_non_define.cas
dburgener: Actually, I think the test as written is wrong here.  What I wrote in the doc is:

> Unlike concrete types, virtual types inherit virtual functions from their parents.

So in this case, the virtual should be automatically inherited.  However, for a *concrete* type, the function would need to be explicitly declared or derived.  I think this is all spelled out clearly in doc/TE.md "Type Inheritance" section, but to try to recap concisely and cleanly here:

* A concrete type should inherit its parents' concrete functions automatically, and a virtual type should inherit its parents' virtual functions automatically.  The exception to both of these is a name conflict.
* For virtual functions not automatically inherited, the child must provide an implementation either explicitly or using the @derive macro.
* For concrete functions, the child may provide an implementation either explicitly or using the derive macro, regardless of whether they were automatically inherited.

In light of that, I believe this test is incorrect in declaring "foo" as virtual, and will break once the above is fully implemented, so I'll update it to remove the virtual keyword.

src/parser.lalrpop
dburgener: I've gone back and forth on that a few times.  I think an argument in favor of the empty `{}` syntax is that that's the expected syntax for implementing a type with no specific rules:

```
resource foo inherits some_parent {}
```

Doing it this way creates a clean "declarations are always blocks" and "statements end in semicolon" distinction.  I see the "like C and rust" argument, so I'm open to pushback, but I do think this way is more internally consistent.  I think declaring virtual functions with a semicolon, becomes an exception that needs to be memorized.  It's a logical an obvious exception, but an exception none the less.

data/error_policies/virtual_function_non_define.cas
```diff
@@ -2,4 +2,4 @@ virtual resource foo_parent {
 	virtual fn foo_func(domain source) {}
 }
 
-virtual resource foo inherits foo_parent {} // Illegal non-inheritance
+resource foo inherits foo_parent {} // Illegal non-inheritance
```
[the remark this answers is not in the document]
dburgener: This is related to the above discussion.  Automatic inheritance of functions outside of resource association isn't implemented yet.  I'm planning to implement that along with the @derive annotation.  Once I do that, the "virtual" version of this would compile successfully (and I'll add a positive test for it then)

data/error_policies/virtual_function_non_define.cas
dburgener: I haven't implemented the function derivation portion yet, so most of this isn't working.  I'd like to add the full set of test cases along with the @derive annotation which I'll be working on in June.

l0kod (approved):

PR #18, opened 2022-04-28
Fix resource association example in docs

dburgener:
The previous example misused the "this" keyword to refer exclusively to
the original type.  Fix it to use the type name directly and add a note
clarifying the distinction.

### DIFF
--- a/doc/resource_association.md
+++ b/doc/resource_association.md
@@ -73,7 +73,7 @@ resource tmp_file {
 		this.manage_files(source);
 		this.manage_dirs(source);
 		this.manage_symlinks(source);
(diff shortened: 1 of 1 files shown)


l0kod (approved):
We should think about testing doc code snippets. I'm not sure how difficult it would be though.

dburgener:
> We should think about testing doc code snippets. I'm not sure how difficult it would be though.

I agree 100%.  I think we need some additional compiler support to make that work though, since they often rest on assumptions of things defined elsewhere.  Let me add an item to the schedule to make this happen later down the road.

PR #20, opened 2022-05-03
Add support for named arguments.

dburgener:
This allows specifying arguments by name (name=value syntax).

It also adds preliminary support for default values, but that will be
fully implemented in a follow-up PR.

Here's the story on switching the built-in function structs to owned
types:

Previously, validating an argument didn't need to take data from
definition of the argument to store the validated argument.  This was
good, because for built in functions we create fake arguments which
don't live past the definition of the function, so we can't keep any
reference to them for later (outputting the final policy).

When we add default value support, now we may have some data flow from
the argument definition to the validated argument (namely, the default
value).  In our current policy, we don't define default values for
built-ins, but:

1. I expect that to change in the future.  The resource_transition()
built-in will have a default name field of None/NULL/the empty
string/TBD.
2. AFAIK even without default values in built ins, there's not a clean
way to make the lifetimes work, since it's not possible for the compiler
to prove in the general case that we never use default values for
built-ins.

That means that our built-in functions, which don't live long enough,
flow data to validated arguments.  That means that those validated
arguments need to not store references to the original data.  The main
place that was happening was built-in functions.  Changing all of those
to owned types and creating copies allows the ephemeral function
arguments for built in functions to work with lifetimes.


data/policies/named_args.cas
```diff
@@ -0,0 +1,15 @@
+domain some_domain {
+	fn three_args(resource a, resource b, resource c) {
+		allow(this, a, file, write);
+		allow(this, b, file, read);
+		allow(this, c, file, open);
+	}
+
+	// TODO: It looks like using "this" as a class name is broken
```
l0kod, line 8: Not good :/

src/internal_rep.rs
```diff
@@ -443,16 +443,16 @@ pub enum AvRuleFlavor {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct AvRule<'a> {
```
l0kod, line 446: Why removing these references?

data/policies/named_args.cas
dburgener: Not great, no.  But it should be a pretty easy fix.  I'll have a PR for it later this month.

And that's just sort of incidental to this PR.  It came up because I wanted to use `this` in a class_name context as part of this PR.  So I think we can go ahead with this PR for now, and when I fix the bug mentioned here, I'll also convert this test to take advantage of it.

src/internal_rep.rs
dburgener: I think I explained this in the commit message.  Let me know if you're looking for more clarification.

dburgener: I've just discovered that Cow may actually give me exactly what I need here.  Working on an update to that.

dburgener: I've just pushed an update to switch all this to Cows, which seems to work well.  Curious what you think, particularly around putting the Cows in the Context::new() definition.  It's a little annoying from an encapsulation standpoint, but the only place new is used directly outside try_from() is test code, and making the Cow owned vs borrowed decision in try_from() seems natural to me.

l0kod: I agree that it is a bit verbose, but it looks good to me. You can squash this last commit with the previous one though.

l0kod (approved):

PR #23, opened 2022-05-12
Allow using the "this" keyword in a function call context

dburgener:


dburgener:
Leaving this as a draft for now because I don't have a test for it.  I noticed this issue in #20, so updating that test to use the "this" keyword as intended should test this well.  #16 also touches the same portion of the code, and this could use a minor update once it merges.  Once those two merge, I'll rebase this, add the test and publish.

dburgener:
Rebased on #16 and added the "this" resolution to alias resolution as alluded to above.  Still waiting on #20 to publish.

src/internal_rep.rs
```diff
@@ -1702,6 +1704,22 @@ impl ValidatedCall {
     }
 }
 
+// If the class_name is "this", return the parent type name,
+// else return the class name.  If the class_name is "this", and the parent type is None, that is
+// an internal error
```
l0kod, line 1709: Would it be possible to name/declare a resource "this"? What forbids it?

dburgener: Good point.  We should probably add a list of reserved keywords and ban declarations with that name.  I'll add that to the backlog.  In my opinion we shouldn't let that block this, since that's broken everywhere for all keywords.

l0kod (approved):

PR #30, opened 2022-06-03
Add local symbol binding

dburgener:


dburgener:
There's work to be done here, but since this is a lot already, I wanted to get at least this much up for review.  See "Future work" comment in the second commit message for the two main things that are left to do.

l0kod (approved):

PR #32, opened 2022-06-07
Implement "extend" keyword

dburgener:
This keyword allows you to extend a type declared elsewhere.

It is an error to use the extend keyword on an undeclared type.

One minor note is that this moves aliases before associations in the
compilation order.  This is so aliases can be extended.  I don't *think*
there's any reason that associations need to be done first.


data/policies/extend.cas
```diff
@@ -0,0 +1,15 @@
+resource foo {
+	allow(bar, foo, file, getattr);
```
l0kod, line 2: Could you add something using `this` like: `allow(bar, foo, this, setattr);`?

data/policies/extend.cas
```diff
@@ -0,0 +1,15 @@
...
+}
+
+extend foo {
+	fn my_func(domain source) {
+		allow(source, this, file, read);
+	}
+
+	allow(bar, foo, file, write);
```
l0kod, line 10: Same for the `this` use too.

data/policies/extend.cas
```diff
@@ -0,0 +1,15 @@
+resource foo {
+	allow(bar, foo, file, getattr);
+}
+
+extend foo {
+	fn my_func(domain source) {
```
l0kod, line 6: I'd like to see a (failed) test with the same function being defined in the original resource and the extended one.

l0kod:
Other than these few comments, it's really cool!

data/policies/extend.cas
dburgener: Did you mean to replace "foo" with "this" instead of "file"?  Because using "this" as the third argument would be an error.

I'll have a local fix changing "foo" to "this" where applicable, which I'll push once I've addressed the other comments.  I think that addresses the concern.  Let me know.

data/policies/extend.cas
dburgener: It turns out this is broken regardless of being in an "extend" block or not.  (There's a FIXME note in compile.rs about it).

I'm working on an update to handle it pretty generically, at AliasMap insertion time, but there are quite a few different cases, notably if one of the duplicated names was synthetically created rather than being explicitly declared, we need to take some care in making sure that we write a clear error message to explain to the user why there is a conflict.

So I'm beginning to wonder if it would make sense to handle that bug as a separate task since it could probably use some more focused attention in a PR to make sure all the cases are handled appropriately.

dburgener: Actually, still not quite working.  I'll aim to push a second commit tomorrow AM.

dburgener: Pushed

dburgener:
FYI, the rebase on top of #38 was highly non-trivial, so you may want to re-review the whole thing.

athvu:
This looks good!

athvu (approved):
This looks good!

PR #33, opened 2022-06-09
Switch derived resources created via association to use "." in their names

dburgener:
instead of "-".

This is more consistent with expectations and typical usage.

### DIFF
--- a/src/parser.lalrpop
+++ b/src/parser.lalrpop
@@ -85,7 +85,7 @@ FuncDeclArg: DeclaredArgument = {
 Stmt: Statement = <StmtBody> ";";
 
 StmtBody: Statement = {
-	<c:Symbol> "." <n:Symbol> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(Some(c), n, a))),
+	<c:TypeName> "." <n:NameDecl> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(Some(c), n, a))),
 	<n:Symbol> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(None, n, a))),
 	"let" <n:Symbol> "=" <a:Arg> => Statement::LetBinding(Box::new(LetBinding::new(n, a))),
 }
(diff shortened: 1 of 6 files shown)


src/parser.lalrpop
```diff
@@ -102,7 +102,7 @@ Stmt: Statement = {
 }
 
 StmtBody: Statement = {
-	<c:Symbol> "." <n:Symbol> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(Some(c), n, a))),
+	<c:TypeName> "." <n:NameDecl> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(Some(c), n, a))),
 	<n:Symbol> "(" <a:Comma<Arg>> ")" => Statement::Call(Box::new(FuncCall::new(None, n, a))),
```
l0kod, line 106: It would imply more changes, but would it be possible to merge these two lines with only `<TypeName> "(" `?

src/parser.lalrpop
```diff
@@ -121,6 +121,11 @@ Ann: Annotation = {
 	"@" <s:NameDecl> => Annotation::new(s),
 }
 
+TypeName: CascadeString = {
+	Symbol,
+	<start: @L> <l: TypeName> "." <r: Symbol> <end: @R> => CascadeString::new([l.as_ref(), ".", r.as_ref()].concat(), start..end)
```
l0kod, line 126: Why `Symbol` but not `NameDecl` instead? It would be nice to forbid the use of `BuiltInType` names and other keywords (e.g. "let", "module", "if"…).

src/parser.lalrpop
dburgener: Looking into this, I think the next line is wrong, and should have been `NameDecl` instead of symbol.  Turning the whole thing into `TypeName` has the problem that `TypeName` concats the individual `Symbols`, and we need the final NameDecl separate (everything before the final `.` is a type, anything after is a function name).

However, once I fixed the next line, the whole thing can just simplify to one line with `<c:(<TypeName> ".")?> <n:NameDecl>`, so the cleanup is pretty straightforward and seems correct to me.

I'll push after I look more at your other comment as well.

src/parser.lalrpop
dburgener: I don't think we want to forbid built in types everywhere that TypeName is used.  For example, the definition of List or Argument both refer to TypeName, but I think built in types can work fine in those contexts.

There are probably a subset of cases here where banning built-in types makes sense (ie a TypeName such as `foo.domain`), but I think that that will definitely fail in the actual compilation because it will never exist because it's impossible to declare a resource named `domain` (or if not, it should be and that's the bug to fix).  So maybe we should enforce this more strictly at the parser, but that might be low priority follow-up work, I think.

I think keywords that exist as strings in parser.lalrpop match as strings first in the lexer and therefore can't match Symbol.

l0kod (approved):

PR #34, opened 2022-06-17
Add optional -o argument

athvu:


dburgener (approved):
Looks great, thanks!

dburgener:
@l0kod This makes me realize that we don't really have a great place for documentation on casc itself yet (besides clap's autogenerated --help).  We need to add a casc man page I think.  It looks like clap recently added a feature to generate manpages using the clap output that might be helpful (https://www.reddit.com/r/rust/comments/sngxf5/ann_clap_mangen_v01_manpage_generation_for_your/).  I'll plan on taking a stab at that soon.

PR #35, opened 2022-06-20
Add a basic man page

dburgener:


dburgener:
This is very much a work in progress.  Here are the issues I'm aware of:

1. The output is going to the build dir.  I think it fits best in doc
2. The "About" string describes the cascade language, not the casc compiler.  I think it's just read out of the Cargo.toml.  This is a problem for casc --help as well.
3. There is no custom information in the man page at this point.  A longer description would probably be helpful

@l0kod as we discussed offline last week, it would be nice to release a 0.0.2 prior to LSS this week so we have all the recent work in a release.  I would consider this a "nice to have" for that, but I also want to get it into good shape before including it, and I'm not sure whether that fits into this week.  Any thoughts on any of the above?

dburgener:
Rebased and addressing item @2 above.  Items 1 and 3 are still outstanding.

In terms of item 1, I'm struggling to determine rust best practices in terms of man page output location.  I've done some googling around crates.io and manpages, with no success.  I found this discussion which seems highly inconclusive about what users generating man pages should actually do with them: https://github.com/rust-cli/man/issues/33.  I think my initial take is that maybe the build dir is sensible for now.  It seems like the most reasonable default.  If we can determine what best practices exist, and how to best integrate with crates.io, distro packagers etc, then we can move it if needed.

athvu (approved):

PR #38, opened 2022-07-18
Module blocks

athvu:


src/compile.rs
```diff
@@ -320,6 +324,205 @@ pub fn validate_functions<'a, 'b>(
...
+            }
+            match &created_modules.iter().find(|x| x.name == m.as_ref()) {
+                Some(x) => other_modules.push((**x).clone()),
+                None => (),
+            }
+        }
+    }
+    let name = &curr_module.name;
+    for d in &curr_module.domains {
+        match types.get(d.as_ref()) {
```
dburgener, line 510: I think it's a little strange to call types.get() in the validation, and then again when creating the ValidatedModule.  Can we not save the TypeInfos from when we validate?

src/compile.rs
```diff
@@ -320,6 +324,205 @@ pub fn validate_functions<'a, 'b>(
...
+
+fn create_validated_modules<'a>(
+    curr_module: &'a Module,
+    mut created_modules: Vec<ValidatedModule>,
+    modules_vec: &Vec<(&SimpleFile<String, String>, &'a Module)>,
+    types: &'a TypeMap,
+) -> Vec<ValidatedModule> {
+    let mut other_modules: Vec<ValidatedModule> = Vec::new();
+    let mut type_infos: Vec<TypeInfo> = Vec::new();
+    if curr_module.modules.len() != 0 {
```
dburgener, line 487: I think the explicit empty check is unnecessary.  If the len is 0, then the for loop just becomes a noop.

src/compile.rs
```diff
@@ -320,6 +324,205 @@ pub fn validate_functions<'a, 'b>(
...
+                    modules_vec.push((&p.file, m));
+                }
+            };
+        }
+    }
+
+    // Validate that module contents exist
+    for (file, module) in &modules_vec {
+        for domain in &module.domains {
+            match types.get(domain.as_ref()) {
```
dburgener, line 351: The match blocks for domain and resource are nearly identical, and quite long.  I think they need to get pulled out into a function or macro to reduce code duplication.

src/internal_rep.rs
```diff
@@ -1766,6 +1770,27 @@ impl ValidatedCall {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ValidatedModule {
+    pub name: CascadeString,
+    type_infos: Vec<TypeInfo>,
+    validated_modules: Vec<ValidatedModule>,
```
dburgener, line 1777: Rather than having a ValidatedModule own its children, it would be better to introduce a BTreeSet named ModuleMap that owns all the modules, and an individual ValidatedModule should have references to the module children.  Once we start building, we'll be building arbitrary modules in the tree, and the fact that they happen to have had some higher level structure is irrelevant.

src/internal_rep.rs
```diff
@@ -1766,6 +1770,27 @@ impl ValidatedCall {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ValidatedModule {
+    pub name: CascadeString,
+    type_infos: Vec<TypeInfo>,
```
dburgener, line 1776: This should be a Vec<&TypeInfo> rather than owning the TypeInfos.  That get rid of any potential problems if the TypeInfo is later changed, and eliminate the need for the clone() call at compile.rs:517

src/compile.rs
```diff
@@ -320,6 +324,205 @@ pub fn validate_functions<'a, 'b>(
...
+            }
+            None => (),
+        }
+    }
+    ret.into_result(())
+}
+
+fn create_validated_modules<'a>(
+    curr_module: &'a Module,
+    mut created_modules: Vec<ValidatedModule>,
```
dburgener, line 481: This whole function feels like a recursive function that's really trying to be an iterative function.  Each call to create_validated_modules() takes all the modules so far, and then returns a complete list.

I've been going back and forth about a cleaner approach, since I feel like this whole clone/recursion thing will be difficult to maintain.

What if we created the modules initially with just the domains and resources and left the child modules as an empty Vec.  Then do a second pass, where we can just straightforwardly look up the already built ValidatedModules and store references to them in the Vec?  I think that will be a lot clearer and cleaner.

src/internal_rep.rs
dburgener: Basically the same approach as "TypeMap" and "FunctionMap".

dburgener (changes requested):
Good start overall.  I think the fundamental issue here is the ownership around ValidatedModule.  Each module should own references to its children rather than the children themselves.  This has at least 3 benefits:

1. A module may have multiple parents, which I don't think this approach supports
2. Easier look-ups into the middle of the module tree (other code doesn't need to care about parents
3. I think it should simplify the create_validated_modules() function.

I think this change will shift things around enough that I basically stopped reviewing at this point, and I'll look over the whole thing again once that's straightened out.

ROADMAP.md
```diff
@@ -36,7 +36,7 @@ Conditionals - Combining both tunables and booleans into a single feature with c
 Ephemeral object and filesystem labeling implementation
 Module blocks - Defines a collection of types that are collectively grouped into a module
```
dburgener, line 37: This line is being achieved by this PR, so it can be deleted as part of the PR.

data/policies/module_alias.cas
```diff
@@ -0,0 +1,24 @@
...
+	allow(this, foo, file, read);
+	allow(this, bar, file, write);
+	allow(this, baz, file, open);
+	foo.list(this);
+}
+
+@alias(thud)
+domain babble {}
+
+module qux {
```
dburgener, line 20: It looks like you're intermixing tabs and spaces in this block.

dburgener (changes requested):

src/compile.rs
```diff
@@ -320,6 +324,189 @@ pub fn validate_functions<'a, 'b>(
...
+                        Some(module_info.0),
+                        m.get_range(),
+                        "This module contains itself or has a descendent that contains it",
+                    ),
+                ));
+            }
+            let inner_module = module_info.1;
+
+            let mut new_visited_modules = visited_modules.clone();
+            new_visited_modules.insert(module_to_check.name.as_ref());
```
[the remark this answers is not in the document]
athvu: Just tested this case and the cycle is still caught. Line 433 checks if the module includes itself directly. 
`if visited_modules.contains(m.as_ref()) || *m == module_to_check.name {`

src/compile.rs
```diff
@@ -320,6 +324,189 @@ pub fn validate_functions<'a, 'b>(
...
+                        "Cycle detected",
+                        Some(module_info.0),
+                        m.get_range(),
+                        "This module contains itself or has a descendent that contains it",
+                    ),
+                ));
+            }
+            let inner_module = module_info.1;
+
+            let mut new_visited_modules = visited_modules.clone();
```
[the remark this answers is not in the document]
dburgener: Scratch this.  After below discussion I understand the approach here better now.  Clone makes sense.

src/compile.rs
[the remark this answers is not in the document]
dburgener: This makes sense.  Thanks for checking and clarifying.

src/internal_rep.rs
```diff
@@ -39,15 +40,24 @@ pub type AliasMapValuesMut<'a, T> = std::collections::btree_map::ValuesMut<'a, S
...
+        if aliases.contains_key(key) {
+            &aliases[key]
         } else {
             key
-        };
+        }
+    }
+
+    pub fn get(&self, key: &str) -> Option<&T> {
+        let type_name = AliasMap::<T>::get_type_name(&self.aliases, key);
```
dburgener, line 52: I might be wrong, but can't this be shortened to "Self::get_type_name("?

src/internal_rep.rs
```diff
@@ -1766,6 +1780,43 @@ impl ValidatedCall {
...
+pub type ModuleMap<'a> = AliasMap<RefCell<ValidatedModule<'a>>>;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd)]
+pub struct ValidatedModule<'a> {
+    pub name: CascadeString,
+    types: BTreeSet<&'a TypeInfo>,
+    pub validated_modules: BTreeSet<&'a RefCell<ValidatedModule<'a>>>,
+}
+
+impl<'a> Ord for ValidatedModule<'a> {
```
dburgener, line 1792: Looking at docs, I see the requirement that Ord must be consistent with PartialOrd https://doc.rust-lang.org/std/cmp/trait.Ord.html and PartialOrd must be consistent with PartialEq: https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html.  I think the derivations will just compare all elements, and since Ord now only compares names (which we want), they'll be inconsistent.

So I think we need manual implementations for PartialEq and PartialOrd as well.

Then of course, there's Eq: https://doc.rust-lang.org/std/cmp/trait.Eq.html  My reading of the docs there is that we *can* derive it.  It requires that children are Eq as well, but that should be fine, and it implements no methods, so there's no need for consistency.

dburgener (approved):

dburgener:
Merged, thanks!

PR #39, opened 2022-07-26
Implement @derive annotation

dburgener:


src/internal_rep.rs
```diff
@@ -386,6 +392,23 @@ impl TypeInfo {
         }
         false
     }
+
+    pub fn get_all_parent_names<'a>(&'a self, types: &'a TypeMap) -> Vec<&'a CascadeString> {
+        let mut ret = Vec::new();
```
l0kod, line 397: Why not use a `BTreeSet` for `ret`? `derived_body` already use this type and it would make it easier to check for duplicate values.

src/internal_rep.rs
```diff
@@ -1549,6 +1687,73 @@ impl<'a> FunctionInfo<'a> {
...
+                None => continue,
+            };
+
+            // All parent functions must have the same prototype.  If this is the first function we
+            // are looking at, we save that prototype.  Otherwise, we compare to ensure they are
+            // identical
+            match first_parent {
+                None => first_parent = Some(parent_function),
+                Some(first_parent) => {
+                    if parent_function.args != first_parent.args {
```
l0kod, line 1723: I guess in a future evolution, Cascade could allow same argument types but with different names.

l0kod (commented):
This looks good overall. I reviewed until the annotation commit but here is a first batch of comments.

src/internal_rep.rs
```diff
@@ -1747,7 +1755,8 @@ impl<'a> FunctionInfo<'a> {
             original_body: &[], // Unused after validation
             body: Some(derived_body),
             declaration_file: file,
-            is_associated_call: false, // I *think* the behavior should be that this is the true if it's true for *any* parent.  TODO: check/document that
+            is_associated_call: derived_is_associated_call,
```
l0kod, line 1773: Hmm, I'm not sure but I think it would make more sense to consider `@associated_call` as part of the function signature and then forbid function composition if they don't all have this annotation.

data/policies/derive.cas
```diff
@@ -34,3 +43,9 @@ domain some_domain {
 	strategy_foo.read(this);
 	custom_define.read(this);
 }
+
+@derive([some_associated_call], strategy=union)
+virtual resource to_associate inherits foo, bar {}
+
+@associate([to_associate])
```
l0kod, line 50: I guess there should be a resulting `call associates-to_associate-some_associated_call` because of the `@associate` annotation, right?

src/internal_rep.rs
```diff
@@ -1721,12 +1721,27 @@ impl<'a> FunctionInfo<'a> {
...
+                            first_parent.get_declaration_range(),
+                            parent_function.get_declaration_range(),
+                        ) {
+                            (Some(first_range), Some(second_range)) => {
+                                return Err(CompileError::new(
+                                        &format!("In attempting to derive {}, parent functions do not have matching prototypes.", name),
+                                        first_parent.declaration_file,
+                                        Some(first_range),
+                                        "This parent prototype...",
+                                        ).add_additional_message(
```
l0kod, line 1734: Nice!

src/compile.rs
```diff
@@ -332,6 +332,11 @@ pub fn validate_functions<'a>(
             Ok(_) => (),
             Err(e) => errors.append(e),
         }
+    }
+
+    derive_functions(&mut functions, types, class_perms)?;
```
l0kod, line 337: What this `fixup! Implement the core of the @derive annotation` commit fix? Could you add a test to prevent this kind of future regression?

data/policies/trait.cas
```diff
@@ -10,8 +10,9 @@ resource foo inherits my_trait {
 	}
 }
 
+@derive([write], strategy=union)
```
l0kod, line 13: This line is correct but not required, right? If we remove the `@derive` annotation we should get the same CIL, correct? What this test is actually testing?

l0kod (commented):
Thanks for splitting these commits!

src/internal_rep.rs
```diff
@@ -386,6 +392,23 @@ impl TypeInfo {
...
     }
+
+    pub fn get_all_parent_names<'a>(&'a self, types: &'a TypeMap) -> Vec<&'a CascadeString> {
+        let mut ret = Vec::new();
+        for parent in &self.inherits {
+            if !ret.contains(&parent) {
+                ret.push(parent);
+                if let Some(parent_ti) = types.get(parent.as_ref()) {
+                    for name in &parent_ti.get_all_parent_names(types) {
+                        if !ret.contains(name) {
```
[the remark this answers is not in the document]
dburgener: Because of the recursive call, ret may contain types multiple levels of inheritance up.  Simple example would be:

```
resource foo {}

resource bar inherits foo {}

resource baz inherits foo {}

resource qux inherits bar, baz {}
```

In this case, a call to foo.get_all_parent_names() will yield foo twice, once recursively calling bar, and once recursively calling baz.  Since this is a totally reasonable thing to do, we just want to dedup, rather than error.

src/compile.rs
dburgener: This is tested by the test added in a3247c0abd2c6c9e37687cda9e6848fa821722d6.  Adding that test exposed the bug.  I intended to squash the fixup commit into the original prior to pushing, but forgot, so it will be squashed on the next push.

src/internal_rep.rs
dburgener: I know, right?  The multi-code point messages were added in #40, and I think they'll come in really handy as we keep developing this.

data/policies/trait.cas
dburgener: This test is testing #51.  Since derive support hadn't merged when we merged #51, the test only including explicitly implemented children, and this commit addresses a TODO there to test compatibility with the @derive feature.

The line is required.  The CIL with the derive annotation includes the line:

```
(macro baz-write ((type this) (type source)) (allow source this (file (write))))
```

which is generated by this derive.

If we remove the `@derive` annotation, it fails to compile with this error:

```
error: baz does not define a function named write
   ┌─ data/policies/trait.cas:13:23
   │
13 │ resource baz inherits my_trait {}
   │                       ^^^^^^^^ All types inheriting my_trait are required to implement write because it is marked as virtual

```

Which I now see has a wrong error message, since "is marked as virtual" is inaccurate.  It should be an error, but the message should say something about a trait instead.  I'll look into that, I thought I remembered fixing it a couple of weeks ago.  Maybe it's on another branch...

Regardless of the bad message, my_trait is a trait, which requires the implementation of the child function write, and this `@derive` is how baz is doing that.

dburgener:
@l0kod Thanks for the review!  I've addressed the low hanging fruit only at this point.  I need to think more on the syntax and associate related bits and formulate a proper response.  I'll follow up later this week.

src/internal_rep.rs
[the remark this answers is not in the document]
l0kod: OK, makes sense. I don't remember if explicitly listing duplicated inherited resources raised an error though.

data/policies/trait.cas
l0kod: Ok, but I though because the `my_trait` has a default implementation it would not be required for `baz` to explicitly add a derive annotation. Do we really want to require such explicit annotation for a default implementation? I would not expect so, as well as if `baz` inherited `foo`.

data/policies/derive.cas
```diff
@@ -0,0 +1,29 @@
...
+	}
+}
+
+resource custom_define inherits foo, bar {
+	fn read(domain source) {
+		allow(source, this, lnk_file, read);
+	}
+}
+
+@derive([read], strategy=union)
```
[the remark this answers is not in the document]
dburgener: I feel like the "empty list" approach seems like just as much of an exception as adding a keyword for it.  I do hear your argument that it's weird that this is sometimes a type name and sometimes a keyword, but I don't really see using [] as any better.  I think the average person encountering that for the first time would be more likely to assume it meant a noop function rather than unioning all.

What if we used one `all` keyword for both strategy and function, and changed `strategy` to union?  So you could say something like `@derive([all], [all])`.  That would at least be more internally consistent.

> Or maybe interpret no explicit strategy as an union?

Well, that already happens because "union" is the default value of the argument.  But there should still be some way to refer to it explicitly.

Another factor here is that we've discussed a few times the possibility of someday having other strategies for how derivation works.  I don't have any particular ideas for what strategies we might employ, but I'm a little hesitant to define a syntax now that makes adding mores strategies later in a backwards compatible way more difficult.

src/compile.rs
```diff
@@ -359,7 +362,61 @@ pub fn validate_functions<'a, 'b>(
...
+        class_perms,
+        target_type.declaration_file.as_ref().unwrap(),
+    )?;
+
+    let parents = match &strategy {
+        DeriveStrategy::Union => target_type.get_all_parent_names(types),
+        DeriveStrategy::Parent(parent) => vec![parent],
+    };
+
+    if vec![CascadeString::from("all")] == func_names {
```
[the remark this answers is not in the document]
dburgener: Can you be more explicit about what you don't like about it?  Languages have reserved keywords and you can't name things them.  I don't know that the inability to name a function "all" is a huge problem?

I'm not necessarily opposed to "*", but it does feel a little terse, and I'm not sure I see a big argument against "all".  On the empty list, see my response on union.

src/internal_rep.rs
dburgener: I believe that is the case in this PR.  See `impl PartialEq for FunctionArgument`.

src/internal_rep.rs
dburgener: Yeah, I think you're right.  I'll see about changing that for the next version of the PR.

data/policies/derive.cas
dburgener: Yes....

It looks like the current CIL has `(call to_associate-some_associated_call (to_associate associates))`, which I think is wrong.  The function name `associates-to_associate-some_associated_call` is neither declared or called, and I think it should be.  Let me look into that some more and see what's going on.

data/policies/trait.cas
dburgener: I think that the idea behind the default implementation was that it enables deriving.  Without such a default implementation you couldn't derive at all.

I think explicit is better when we're talking about security.  The hope with the `@derive` annotation is to allow developers ease of use without taking away control and having too many surprising side effects under the hood.  I think its consistent with rust and works well (some rust traits are derivable, and it's straightforward to derive them, but you can also provide your own implementation).

The idea here is that with inheriting my_trait, you are promising that you will provide some `write()` function matching the prototype.  It's up to the developer whether they want to just take the default, provide their own, or perhaps union several versions together with multiple inheritance, but we do require them to be explicit about it.  That way it's a conscious decision and if they don't think about it the compiler doesn't just decide to do something they may not expect, it prompts them to make a choice.

Also, as you mention with multiple inheritance, a default is really non-intuitive.  If `baz` inherited both `foo` and `my_trait`, does the fact that `my_trait` is a trait mean that the default should just be `my_trait`'s write?  Or should the default include a union with `foo` in that case?  I feel like people might want either, so it's better to be explicit.

src/internal_rep.rs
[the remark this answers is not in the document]
dburgener: Just tried it out, and it compiled fine.  I'll file an issue for later cleanup.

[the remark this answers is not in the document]
dburgener: filed #56

src/internal_rep.rs
l0kod: Indeed. What append if arguments have the same type but not the same name? I don't see a "name conversion" nor related tests.

dburgener: Yeah, good catch.  I suspect we're broken in that situation.  I'll update for the next push.

I expect the exact name doesn't matter in that situation, just that it needs to be consistent.  So we should be able to simply choose a name from one of the functions (or come up with our own name that we know is invalid elsewhere).  Since the name used is internal to this function and only appears in compiled output.

I guess the question there would be if it ever shows up in error messages or future analysis/diagnostic tools.  Maybe something consistent like "_arg1", "_arg2" would be sensible across the board in all derived functions?

l0kod: Choosing the right name generation is indeed worth thinking about because of error messages. I think we should try to keep as much as possible the original names, maybe concatenating them in a deterministic way (`sort -u`-like with `_` in between)?

dburgener: I think that seems reasonable.  I've pushed a new commit that implements that.

data/policies/derive.cas
dburgener: Alright, there were a couple of interacting bugs down there, so it was a pain to track down, but I believe I've gotten it all sorted out, and the call you mentioned now appears, with the erroneous one I mentioned gone.

The two bugs were:

* The comparison of which function name to call used the name of the resource being associated rather than the synthetic resource name
* Synthetic resources created by association cleared all annotations, which meant they didn't actually end up deriving after all.  I changed that to have them keep @derive and @makelist, which I think makes sense.  @alias isn't really possible since it defines a unique name that can't refer to two things.  @associate probably wouldn't hurt anything since it's already been processed at that point, but for the same reason, there's no point in keeping it.  For future annotation additions, we'll need to keep that list updated, so it should probably be generalized somehow, but I'll leave that aspect for future enhancements.

With this latest change, I believe I've addressed all your comments, either by fixing something or responding to the comment.  Would appreciate another round of feedback and your responses to my responses when you have time.

src/internal_rep.rs
```diff
@@ -2572,6 +2578,43 @@ mod tests {
...
+        assert_eq!(&fi.get_cil_name(), "foo");
+
+        let ti = TypeInfo::new(
+            TypeDecl {
+                name: CascadeString::from("bar"),
+                inherits: Vec::new(),
+                is_virtual: false,
+                is_trait: false,
+                is_extension: false,
+                expressions: Vec::new(),
```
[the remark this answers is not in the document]
dburgener: I generally like this idea.  I feel like there's not a sensible default for the "name" field though, and a Default trait feels weird when one of the fields doesn't have a sensible default.  I don't relish the idea of trying to track down a "why do we have a type whose name is the empty string?" bug someday.  Maybe it's worth it though to simplify constructing TypeInfos.  So far, both the call sites constructing them and the number of fields on the struct have both grown a fair amount, so simplification there would be nice.

Thoughts on the above?  Is there a more idiomatic approach I'm missing?  Is implementing Default and using `CascadeString::from("")` for the name field less of a problem than I'm making it out to be?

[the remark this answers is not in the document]
l0kod: Why not `CascadeString::default()` instead?

src/compile.rs
```diff
@@ -945,7 +1017,13 @@ fn create_synthetic_resource(
     // Virtual resources become concrete when associated to concrete types
     dup_res_decl.is_virtual = dup_res_decl.is_virtual && dom_info.is_virtual;
     let dup_res_is_virtual = dup_res_decl.is_virtual;
-    dup_res_decl.annotations = Annotations::new();
+    // The synthetic resource keeps some, but not all annotations from its parent.
+    // Specifically, Makelist and derive are kept from the parent
+    dup_res_decl
+        .annotations
+        .annotations
+        .retain(|a| a.name.as_ref() == "makelist" || a.name.as_ref() == "derive");
```
l0kod, line 1025: It would be less error-prone to use a dedicated method for `enum AnnotationInfo`. This way we must define this behavior when adding a new annotation.

src/internal_rep.rs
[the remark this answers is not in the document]
dburgener: Creating `CascadeString::default()` is what I'm arguing against.  I'm not sure when I'd ever want empty CascadeStrings, so creating the opportunity to accidentally make them feels weird.

src/compile.rs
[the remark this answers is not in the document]
dburgener: I think I'm tracking with you on the type weirdness.  Certainly the list of `[all]` feels weird.

I think I'm okay with your suggestion of `*` as an alternative.  Something like `@derive(*, *)`.

The `function` argument can be a list of names of functions to derive, or `*` to derive all functions.  Likewise the `union` ("strategy" in the current incarnation) could be either a list of parent names to derive from, or `*` to derive from all parents.

Does that approach work for you?

src/lib.rs
```diff
@@ -796,7 +796,21 @@ mod tests {
...
+    }
+
+    #[test]
+    fn derive_test() {
+        valid_policy_test("derive.cas", &["(macro strategy_union-read ((type this) (type source)) (allow source this (dir (read))) (allow source this (file (read))))",
+        "(macro strategy_foo-read ((type this) (type source)) (allow source this (file (read))))",
+        "(macro custom_define-read ((type this) (type source)) (allow source this (lnk_file (read))))",
+        "(macro derive_all-read ((type this) (type source)) (allow source this (dir (read))) (allow source this (file (read))))",
+        "(macro derive_all-write ((type this) (type source)) (allow source this (dir (write))))",
+        "call associates-to_associate-some_associated_call"],
```
l0kod, line 812: Is it on purpose to not check the full line (with parenthesis)?

data/policies/derive.cas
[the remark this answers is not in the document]
l0kod: What about `derive_union()`? Do you think other strategy would make sense?

src/compile.rs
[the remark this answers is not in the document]
dburgener: @l0kod I was hoping you could respond to this idea?

data/policies/derive.cas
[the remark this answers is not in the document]
dburgener: I don't see what `derive_union` achieves?  It just seems more verbose to me.

src/compile.rs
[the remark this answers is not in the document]
l0kod: The `@derive(functions=*, parents=*)` proposal looks good! The `strategy` argument could be added in the future if we need something different than unions.

src/internal_rep.rs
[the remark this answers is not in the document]
dburgener: @l0kod still need your response on this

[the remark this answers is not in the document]
l0kod: What is the difference between `default()` and `from("")`?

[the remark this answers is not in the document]
dburgener: I don't see a difference.  My point is that having empty CascadeStrings doesn't sound like something we would ever want.

data/error_policies/derive_diff_associated_call.cas
```diff
@@ -0,0 +1,18 @@
...
+	}
+}
+
+virtual resource bar {
+	fn some_associated_call(domain source) {
+		allow(source, this, dir, add_name);
+	}
+}
+
+@derive([some_associated_call], strategy=union)
```
l0kod, line 14: Could we remove the strategy argument for now? It would be handy for other strategies but there is only `union` for now, so I think it doesn't make sense to keep the `strategy` argument.

data/error_policies/derive_non_matching_parents.cas
```diff
@@ -0,0 +1,14 @@
...
+	}
+}
+
+virtual resource bar {
+	fn do_something(domain source) {
+		allow(source, this, file, read);
+	}
+}
+
+@derive([do_something], union)
```
l0kod, line 13: I guess this `union` argument will also go away? It can only confuse users.

data/policies/derive.cas
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+resource custom_define inherits foo, bar {
+	fn read(domain source) {
+		allow(source, this, lnk_file, read);
+	}
+}
+
+@derive([read], parents=*)
```
l0kod, line 33: The `parent=*` named argument is optional right? Can we add a case without this argument?
If it is not optional, why do we need to use a named parameter?

data/error_policies/derive_diff_associated_call.cas
dburgener: It's already been removed.  I just missed removing it in this error test.  Will push an update.

data/error_policies/derive_non_matching_parents.cas
dburgener: Yep.  Same as above.  I missed an update in the error tests.  Thanks for catching it

data/policies/derive.cas
dburgener: Yeah, good call.  I've pushed an update adding this test.

src/internal_rep.rs
```diff
@@ -1770,6 +2127,15 @@ impl From<&FunctionArgument<'_>> for sexp::Sexp {
     }
 }
 
+// Two arguments are equal if their types are the same (including is_list_param).
+// Names and defaults may differ
+impl PartialEq for FunctionArgument<'_> {
+    fn eq(&self, other: &FunctionArgument) -> bool {
+        &self.name == "this" && &other.name == "this"
```
l0kod, line 2134: Is it really needed to add this exception with `this`?

l0kod (approved):
This big PR looks good to me! I added some nitpicking reviews but you can address that in a later PR. You could squash some commits but I guess it would probably require to resolve a lot of conflicts, which would not be worth it.

src/internal_rep.rs
dburgener: This is because all functions internally have a "this" argument inserted as their first argument.  That argument has the type of the parent function.  Since in deriving "this" becomes a reference to the child type, but in its original context it referred to the parent type, these need to be handled as a special case.

PR #44, opened 2022-08-01
parse and validate systems

athvu:


src/compile.rs
```diff
@@ -479,6 +478,145 @@ fn validate_module_contents<'a>(
...
+                        &format!("Module {} does not exist", m.as_ref()),
+                        Some(file),
+                        m.get_range(),
+                        "modules within systems must be declared elsewhere",
+                    ),
+                )),
+            }
+        }
+        // Validate the system's configurations
+        for c in &system.configurations {
```
dburgener, line 519: I think I'd rather that this whole chunk be made more generic.  Like a function that takes a configuration name, a list of valid values, and a bool telling whether it's required, and then either inserts, or returns an error.

src/compile.rs
```diff
@@ -479,6 +478,145 @@ fn validate_module_contents<'a>(
...
+                    None,
+                    &format!(
+                        "Add a {} configuration to system {}",
+                        constants::HANDLE_UNKNOWN_PERMS,
+                        system.name
+                    ),
+                ),
+            ));
+        }
+
```
dburgener, line 611: What about multiple copies of the same config?

system foo {
let monolithic = true;
let monolithic = false;
}

src/internal_rep.rs
```diff
@@ -1802,6 +1820,29 @@ impl<'a> ValidatedModule<'a> {
...
+    pub name: CascadeString,
+    pub modules: BTreeSet<&'a ValidatedModule<'a>>,
+    pub configurations: BTreeMap<String, &'a LetBinding>,
+}
+
+impl<'a> ValidatedSystem<'a> {
+    pub fn new(
+        name: CascadeString,
+        modules: BTreeSet<&'a ValidatedModule<'a>>,
+        configurations: BTreeMap<String, &'a LetBinding>,
```
dburgener, line 1836: String to LetBinding?  Why not String to Argument?

src/lib.rs
```diff
@@ -723,6 +722,11 @@ mod tests {
         error_policy_test!("module_cycle.cas", 1, ErrorItem::Compile(_));
     }
 
+    #[test]
+    fn invalid_system_error() {
+        error_policy_test!("system_invalid.cas", 5, ErrorItem::Compile(_));
```
dburgener, line 727: Can we add invalid tests for:

1. Missing a required config
2. Invalid module
3. Same config multiple times

dburgener (changes requested):

dburgener:
Can you add a little more detail into the commit message?  It doesn't need to be crazy thorough, but another sentence or two briefly explaining what a system is and what this will enable in the future would be helpful for git log readers.

src/ast.rs
```diff
@@ -8,6 +8,7 @@ use std::ops::Range;
 use codespan_reporting::files::SimpleFile;
 
 use crate::constants;
+use crate::ParseErrorMsg;
```
dburgener, line 11: I don't know if it actually matters functionally, but can this be "crate::error::ParseErrorMsg" instead please?

src/ast.rs
```diff
@@ -544,6 +547,75 @@ impl Virtualable for Module {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct System {
+    pub name: CascadeString,
+    pub is_virtual: bool,
```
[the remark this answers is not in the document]
dburgener: Now that set_virtual() can return an error, I think we can just drop this field entirely.

src/ast.rs
```diff
@@ -549,11 +555,84 @@ impl Module {
...
+                SystemBody::Mod(m) => {
+                    self.modules.push(m);
+                }
+                SystemBody::Config(l) => {
+                    self.configurations.push(l);
+                }
+            }
+        }
+        // Insert the default configurations if they were not provided
+        if !self
```
dburgener, line 596: I'd rather have this just loop over a list of [(config_name, default_value), ...].  Similar to how the get_built_in_types_map() function works.  You can even define the default in constants.rs like with get_built_in_types_map().

src/compile.rs
dburgener: I think we can push the code duplication reduction even farther, using the same approach described above in set_fields().

src/compile.rs
```diff
@@ -514,6 +513,189 @@ fn validate_module_contents<'a>(
...
+                Some(file),
+                config.name.get_range(),
+                "Each configuration can only be included once in each system",
+            ),
+        ))
+    } else if let std::collections::btree_map::Entry::Vacant(e) =
+        configs.entry(config_name.to_string())
+    {
+        if let Argument::Var(a) = &config.value {
+            // mls and mcs are not yet supported
```
dburgener, line 652: This comment is particular to policy type and can be removed here

src/error.rs
```diff
@@ -155,17 +162,17 @@ impl From<LalrpopParseError<usize, Token<'_>, &str>> for ParseErrorMsg {
                 help: String::new(),
             },
             LalrpopParseError::User { error } => ParseErrorMsg {
```
dburgener, line 164: Can't you just return `error` directly rather than creating a new ParseErrorMsg that is an exact copy?

src/parser.lalrpop
```diff
@@ -36,11 +42,15 @@ pub Expr: Expression = {
 
 BaseExpr: Expression = {
 	// TODO: other keywords beside virtual?
-	<v: "virtual"?> <mut d: Decl> => {
+	<start: @L> <v: "virtual"?> <mut d: Decl> <end: @R> =>? {
```
dburgener, line 45: I think the error message will look better if we just put the range around the `virtual` keyword instead of the whole decl.

src/compile.rs
```diff
@@ -514,6 +513,189 @@ fn validate_module_contents<'a>(
...
+            }
+        }
+    }
+
+    for (file, system) in &systems_vec {
+        let mut system_modules = BTreeSet::new();
+        let mut configs = BTreeMap::new();
+
+        // Validate that the modules of a system exist
+        for m in &system.modules {
```
dburgener, line 538: I feel like declaring a system with no modules should be an error.  Are we checking that anywhere?

dburgener (changes requested):

doc/modules.md
```diff
@@ -10,4 +10,16 @@ made virtual, in which case it cannot be built. The purpose of a virtual module
...
+
+# Systems
+Systems contain modules and configuration options. Some configurations are
+mandatory, meaning that they must be explicitly included in a system, while
+other are optional. If the user does not include the optional configurations in
+a system, the default values will be given.
+
+Configuration | Mandatory? | Description
+--------------|------------|-------------
+handle_unknown_perms|Yes|This is how permissions missing from the policy will be handled when loading the policy. The options are allow, deny, and reject. [handleunkown](https://github.com/SELinuxProject/selinux/blob/master/secilc/docs/cil_policy_config_statements.md#handleunknown)
```
dburgener, line 23: Typo in link text.  "unkown" -> "unknown"

src/ast.rs
```diff
@@ -549,11 +555,72 @@ impl Module {
...
+                }
+                SystemBody::Config(l) => {
+                    self.configurations.push(l);
+                }
+            }
+        }
+
+        // Insert the default configurations if they were not provided
+        for (config_name, default_value) in constants::SYSTEM_CONFIG_DEFAULTS {
+            let config_name = config_name.to_string();
```
dburgener, line 596: Is this conversion actually necessary?  I believe that PartialEq is defined on &str and String and that From<&str> is defined for CascadeString.  So I'd think that we could just use a &str everywhere in this loop without converting to String first?

data/policies/systems.cas
```diff
@@ -0,0 +1,7 @@
+system name {
```
dburgener, line 1: Is this actually being consumed anywhere?  I think we may need to add a test that this builds.

If you use valid_policy_test(), secilc will complain that all policies must contain an allow rule, so you would need to add a domain or resource into the module and an allow() call into that in order to make this a valid policy.

It would be good to verify the configs and the policy name (which the policy name isn't checkable via valid_policy_test(), so that will need special handling.

dburgener (approved):

dburgener:
merged, thanks!

PR #47, opened 2022-08-19
Implement system building

athvu:


src/lib.rs
```diff
@@ -32,13 +32,21 @@ lalrpop_mod!(#[allow(clippy::all)] pub parser);
...
+/// The dummy system is used to build if no system names are given.
 /// Returns a Result containing either a string of CIL policy which is the compiled result or a
 /// list of errors.
-/// In order to convert the compiled CIL policy into a usable policy, you must use secilc
-pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::CascadeErrors> {
+/// In order to convert the compiled CIL policy into a usable policy, you must use secilc.
+pub fn compile_system_policies(
+    input_files: Vec<&str>,
+    mut system_names: Vec<String>,
+    dummy_system: Option<ValidatedSystem>,
```
dburgener, line 43: I don't think we want to expose ValidatedSystem outside the library.  It looks like casc calls this as None anyways, and the current use case is just for compile_all_system_policies(), right?

The issue here is that compile_system_policies() is `pub`, so it gets exposed to users, and should therefore have a really clean interface.

Off the top of my head, I think the cleanest thing might be to have compile_system_policies() just take input files and system names, and then directly call into a new function compile_system_policies_internal(), which does what compile_system_policies does now.  compile_all_system_policies() can then also call into the _internal() version.

If system_names is empty, we can return an error.

I think "default" is probably a better name than "dummy".

src/lib.rs
```diff
@@ -189,9 +160,102 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::Ca
...
+}
+
+/// Compile all systems into a single policy
+/// The list of input files list should contain filenames of files containing policy to be
+/// compiled.
+/// The system name is the name of the output file (out.cil by default).
+/// Returns a Result containing either a string of CIL policy which is the compiled result or a
+/// list of errors.
+/// In order to convert the compiled CIL policy into a usable policy, you must use secilc.
+pub fn compile_all_system_policies(
```
dburgener, line 241: The naming here is weird.

There are 3 cases, right?

* casc provides a list of system names to build
* casc wants to build everything into one system
* casc wants to build all defined systems

This function name sounds like the third one, but I think it implements the second.

I think the API should be a separate call for each of the three above cases.  Then we have a private compile_system_policies_internal() that does all the actual work.  Each of the three public functions just get system stuff set up appropriately for the internal() function.

src/error.rs
```diff
@@ -199,6 +199,21 @@ impl ParseError {
     }
 }
 
+#[derive(Error, Clone, Debug)]
+#[error("{diagnostic}")]
+pub struct BadSystemError {
```
dburgener, line 204: How about "Invalid" instead of "Bad"?

dburgener (changes requested):
Sorry, I haven't had time to dive into all the details yet.  This is just a quick review on the public API portion of it.

src/lib.rs
athvu: In response to "If system_names is empty, we can return an error," I want to note that with the current implementation, since -s has a default value, there is no error for when the user does not provide system names.
I assume this will be changed, so that -s will no longer have a default value, after we figure out how we want to specify building all systems. However, in that case, clap has its own error for not supplying a value.

dburgener:
The first commit message is still wrong.  That work was done in a previous PR.  Please update the first commit message to match the work actually done by that commit.

src/compile.rs
```diff
@@ -682,6 +706,145 @@ fn check_required_config<'a>(
     ret.into_result(())
 }
 
+pub fn get_infos<'a>(
```
dburgener, line 709: This function seems kind of weird to me.  It looks like it does all the compilation for an individual system except for types.  Which is a slightly strange distinction to make.  I think I'm struggling to interpret and understand this function because the intent is not clear to me.  Perhaps with a more descriptive function name and a comment explaining what this function does it will make sense?  I'm more inclined to think that it does several mostly unrelated things and should either be split up, or it should include the entirety of system building (including types).

src/compile.rs
```diff
@@ -682,6 +706,145 @@ fn check_required_config<'a>(
...
+        &mut new_func_map,
+        new_type_map,
+        classlist,
+        &new_func_map_copy,
+    )?;
+
+    // Get the policy rules
+    let new_policy_rules = get_policy_rules(&new_policies, new_type_map, classlist, &new_func_map)?;
+
+    let new_cil_tree = generate_sexp(
```
dburgener, line 734: Do we have to do this here?  It would be nice if generate_sexp() happened after the system loop so that we can get compile errors on all the systems before getting errors from generate_sexp().

src/compile.rs
```diff
@@ -682,6 +706,145 @@ fn check_required_config<'a>(
...
+        &new_func_map,
+        system_configs,
+    )?;
+
+    ret.into_result(new_cil_tree)
+}
+
+pub fn get_reduced_types(
+    module: &ValidatedModule,
+    reduced_type_map: &mut TypeMap,
```
dburgener, line 747: I'm not a huge fan of this pattern of having a mutable argument and then returning Ok(()).  In this case, I think it makes sense to move the initialization into this function and just return the map in the Ok case.

src/compile.rs
```diff
@@ -682,6 +706,145 @@ fn check_required_config<'a>(
...
+        classlist,
+        new_policy_rules,
+        &new_func_map,
+        system_configs,
+    )?;
+
+    ret.into_result(new_cil_tree)
+}
+
+pub fn get_reduced_types(
```
dburgener, line 745: This could use a comment that goes over the rules for reducing types, notably the fact that we include parents automatically.

It's probably also worth mentioning that we clone the types, because ownership.

src/compile.rs
```diff
@@ -682,6 +706,145 @@ fn check_required_config<'a>(
...
+    }
+    for vm in &module.validated_modules {
+        if let Some(child_module) = module_map.get(vm.as_ref()) {
+            get_reduced_types(child_module, reduced_type_map, type_map, module_map)?;
+        }
+    }
+    Ok(())
+}
+
+pub fn get_policy_files<'a>(
```
dburgener, line 773: Okay, if I understand this right, the assumption here is that:

1. All the files we care about are declaration_files in types in our module.
2. Every function and rule in those files are needed.

Is that accurate?  I'm concerned about this assumption because:

1. A file may contain many types, only some of which are in the system.
2. A type may be "extend"ed using an extend block in some other file.

I think we need to rebrowse all files, but only descend into blocks of included types when generating the later data structure.

src/bin/casc/args.rs
```diff
@@ -8,4 +8,11 @@ pub struct Args {
     pub input_file: Vec<String>,
     #[clap(default_value = "out.cil", short)]
     pub out_filename: String,
+    #[clap(
+        short,
+        multiple_values = true,
+        default_missing_value("all"),
```
dburgener, line 14: So this would imply that I can't declare a system named "all".

I'm fine with that, and actually we're not really good at enforcing that things aren't named reserved keywords yet.

Can you add a comment with this assumption?  I'll file an issue on github to add validation that declarations in general don't include reserved keywords.

dburgener (changes requested):

src/compile.rs
athvu: This comment has not yet been addressed, but I pushed the changes anyway, in case you wanted to take a look at them.
I'm running into errors with returning values referencing local variables that I haven't been able to work out, but will try again tomorrow.

src/lib.rs
```diff
@@ -32,44 +32,69 @@ lalrpop_mod!(#[allow(clippy::all)] pub parser);
...
     }
-    // Stops if something went wrong for this major step.
-    errors = errors.into_result_self()?;
+    compile_system_policies_internal(policies, system_names, false)
+}
+
+/// Compile all systems into a single policy
+/// The list of input files list should contain filenames of files containing policy to be
+/// compiled.
+/// The system name is the name of the output file (out.cil by default).
```
dburgener, line 71: The default is being supplied by casc, not the library, right?  So calling it out here at the casc/library boundary doesn't make sense.  It looks like the code just clones whatever system_name is passed in.  Also, the library doesn't generate a file, it returns the string that is the contents of the file and casc actually makes the file.

I think the code makes sense, and the comment is what need changing.

src/lib.rs
```diff
@@ -32,44 +32,69 @@ lalrpop_mod!(#[allow(clippy::all)] pub parser);
...
+/// In order to convert the compiled CIL policy into a usable policy, you must use secilc.
+pub fn compile_combined(
+    input_files: Vec<&str>,
+    system_name: String,
+) -> Result<String, error::CascadeErrors> {
+    let errors = CascadeErrors::new();
+    let policies = get_policies(input_files)?;
+    let res = compile_system_policies_internal(policies, vec![system_name.clone()], true)?;
+    let ret = res.get(&system_name);
+    errors.into_result(ret.unwrap().to_string())
```
dburgener, line 83: Okay, so ret could be `None` if `compile_system_policies_internal()` returned `Ok(_)`, but didn't store a system named `system_name` in the HashMap.  That would presumably be a bug in compile_system_policies_internal().  I think in the event of that sort of bug we have typically (for better or worse) been returning an InternalError rather than panicing.

dburgener: Also, the `to_string()` is here because HashMap get returns an `Option<&T>`, right?  So ret.unwrap() returns type `&String`?  In that case, I think that just dereferencing with * seems cleaner?  (It looks like to_string() on &String calls to_owned(), which may clone() under the hood, so `*` would avoid the clone).  All this is of course moot if I'm wrong about the types and this is actually a &str for some HashMap reason.

src/lib.rs
```diff
@@ -189,9 +179,86 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::Ca
...
-    let cil_tree = compile::generate_sexp(&type_map, &classlist, policy_rules, &func_map)?;
+    // Create a default module and default system
+    // Insert the default module into the default system and insert the system into the system map
+    let default_module = &mut ValidatedModule::new(
+        CascadeString::from("module"),
+        BTreeSet::new(),
+        BTreeSet::new(),
+        None,
+        None,
+    )?;
```
dburgener, line 190: This isn't going to fail, but if it does, that an implementation bug, so we should return an InternalError rather than passing on the CompileError that I assume ValidatedModule would give us.

src/lib.rs
```diff
@@ -189,9 +179,86 @@ pub fn compile_system_policy(input_files: Vec<&str>) -> Result<String, error::Ca
     // Validate systems
     compile::validate_systems(&policies, &module_map, &mut system_map)?;
 
-    let cil_tree = compile::generate_sexp(&type_map, &classlist, policy_rules, &func_map)?;
+    // Create a default module and default system
+    // Insert the default module into the default system and insert the system into the system map
+    let default_module = &mut ValidatedModule::new(
```
dburgener, line 184: Is there a reason that this isn't inside the `if create_default_system` block?

src/lib.rs
dburgener: I'm looking more at the usages in the tests and in main.rs.  I think there's some weirdness between "output file name" and "system name".

In `compile_system_policies` and `compile_system_policies_all` the system names are the portion of the filename without .cil (eg "out").  In `compile_combined`, the system name is the full file.  Then casc accounts for that by appending ".cil" in the first two cases, but not in compile_combined.

I think the library interface needs to be consistent here.  System names are always just the descriptive name, without ".cil", and the output file name will have ".cil" appended to it.  That means that in casc, the call to compile_combined will need to get the portion of the output filename before .cil, and then append .cil afterwards.

dburgener: Actually, why do we care about this at all?  casc sets the output file name.  Does compile_combined even need to know it?  All compile combined needs is some internal name to use for the system, but that shouldn't end up affecting what the user gets, since we're not actually returning the system name, right?  Can we not just drop the system_name argument here entirely and just use some internal value?

src/lib.rs
```diff
@@ -613,6 +682,51 @@ mod tests {
         }
     }
 
+    #[test]
+    fn system_building_test() {
```
dburgener, line 686: Can we test compile_system_policies_all() as well?

src/compile.rs
```diff
@@ -682,6 +706,131 @@ fn check_required_config<'a>(
...
+        &Some(&system.configurations),
+    )?;
+
+    ret.into_result(new_cil_tree)
+}
+
+// This is a recusive function that gets only the relevant types from the type map.
+// The reduced types are the types in the module and the types in any of that modules' child modules.
+// Parents of those types are also automatically included.
+// Types are cloned for ownership.
```
dburgener, line 758: By "we clone the types because ownership", I more meant something like "Makes a clone of the types so that each system TypeMap can own its own types"

src/compile.rs
```diff
@@ -682,6 +706,131 @@ fn check_required_config<'a>(
...
+    }
+    for vm in &module.validated_modules {
+        if let Some(child_module) = module_map.get(vm.as_ref()) {
+            get_reduced_types(child_module, reduced_type_map, type_map, module_map)?;
+        }
+    }
+    Ok(())
+}
+
+pub fn get_funcs<'a>(
```
dburgener, line 787: Putting this here for lack of a better place...

So, we reduce the function map.  Which means that theoretically, some function (call it foo.my_func) may exist in the policy, but not in the particular system being compiled.  So when I inside bar (which is part of my system) call foo.my_func, then it's no longer in the map.

I'm assuming that in that case, we'd get a "this function doesn't exist message" (hopefully it's that rather than generating invalid policy that calls the function but doesn't declare it?)

If that's all accurate, that's fine for now, but we should capture the current behavior somewhere and make a note while we're in the code and understanding everything how to update it later.  I *think* that in that case, we'll generally be wanting to silently drop those calls rather than returning an error.  That should probably get captured as a TODO for later, since I still need to write up some more formal design for it and discuss with people.

Am I missing anything here?  Can you figure out where to document the status of this particular aspect of compilation and document it please?

src/bin/casc/main.rs
```diff
@@ -34,25 +33,58 @@ fn main() -> std::io::Result<()> {
...
+    // In both of the previous two cases, the name of each output CIL file is the name of the system + .cil.
+    if args.system_names.is_empty() {
+        let mut out_file = File::create(args.out_filename.clone())?;
+        let res = compile_combined(
+            policies.iter().map(|s| s as &str).collect(),
+            args.out_filename,
+        );
+        match res {
+            Err(error_list) => print_error(error_list),
+            Ok(s) => out_file.write_all(s.as_bytes()),
```
dburgener, line 51: Can we unify the writing out code by having the part that calls compile_combined make its own HashMap with the default system name?

dburgener (changes requested):

dburgener:
I don't see any doc updates in this.  Do we need associated documentation changes here?

src/lib.rs
athvu: I tried out using * instead of to_string(), but it doesn't seem like dereferencing a &String is allowed.

dburgener: I played with this a little on my system and this change compiles and passes tests for me:

```
-    let res = compile_system_policies_internal(policies, vec![system_name.clone()], true)?;
-    let ret = res.get(&system_name);
-    errors.into_result(ret.unwrap().to_string())
+    let mut res = compile_system_policies_internal(policies, vec![system_name.clone()], true)?;
+    let ret = res.remove(&system_name);
+    errors.into_result(ret.unwrap())
```

I believe the issue that you're hitting is that the code actually needed the implicit clone() to work from an ownership perspective.  Since res is a `HashMap<String, String>`, ret is a reference into that HashMap.  Trying to take ownership of that reference by dereferencing isn't legal because the HashMap still owns it, so deref() tries to implicitly copy(), which fails because String doesn't impliment the Copy trait.

However, we don't need to retain ownership of the HashMap.  What we really want is to take ownership of the string we'll be returning from the HashMap, and then transfer that ownership to the caller.  Switching the get() call to a remove() call accomplishes that.

doc/policy_composer.md
```diff
@@ -22,4 +22,27 @@ Configuration | Mandatory? | Description
...
 system_type|No|This controls whether MLS/MCS controls are enabled. The only currently supported option is standard. Options for mls and mcs will be added in the future. The default is standard. [MLS/MCS](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/mls_mcs.md)
-monolithic|No|This is the option for choosing to build monolithic or modular policy. The options are true and false, with true being for a monolithic build and false for modular. The default is true. [Monolithic and Modular policies](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/types_of_policy.md#monolithic-policy)
\ No newline at end of file
+monolithic|No|This is the option for choosing to build monolithic or modular policy. The options are true and false, with true being for a monolithic build and false for modular. The default is true. [Monolithic and Modular policies](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/types_of_policy.md#monolithic-policy)
+
+# System Building
+There are 3 options for system building.
+
+(1) compile_combined  
+This is the default when casc is ran without the -s flag. It compiles all
```
dburgener, line 31: All systems, or all defined policy?  In other words, if I have a resource foo and it is included in no modules or systems, it would still be included with compile_combined, right?

dburgener (approved):

dburgener:
Merged.  Thanks!

PR #49, opened 2022-08-31
Add info on casc CLI to manpage

athvu:


src/bin/casc/args.rs
```diff
@@ -1,13 +1,21 @@
 use clap::Parser;
 
 #[derive(Parser, Debug)]
-#[clap(author, version, name = "casc", about = "Compile Cascade SELinux policies into CIL", long_about = None)]
+#[clap(
+    author,
+    version,
+    name = "casc",
+    about = "Compile Cascade SELinux policies into CIL",
+    long_about = "Compile Cascade SELinux policies into CIL\nSupported use cases:\n- Combine all policies: This is the default. \"-o <out_filename>\" to specify the output filename.\n- Build one defined system: \"-s <system_name>\"\n- Build multiple defined systems: \"-s <system_names>\"\n- Build all defined systems: \"-s all\""
```
dburgener, line 9: 1. When I run "man path/to/casc.1", the line splitting doesn't work for me.  Not sure if that's man page formatting or something clap_mangen is doing, but it's all on one line.
2. Can we elaborate on the Combine all policies a bit more?  Like that the output is a monolithic system policy, and that all configuration options are set to defaults
3. I think the "build one" and "build multiple" options can be combined into one here.  A list of systems with one system listed isn't fundamentally different from a list with more.

src/bin/casc/args.rs
```diff
@@ -1,13 +1,25 @@
...
+    - Combine all policies: This is the default. It outputs a monolithic system policy with the system configuration options set to default values. \"-o <out_filename>\" to specify the output filename.
+    - Build one or more defined systems: \"-s <system_names>\"
+    - Build all defined systems: \"-s all\""
+)]
 pub struct Args {
-    /// List of input files to process.  Directories are searched recursively.
+    /// List of input files to process. Directories are searched recursively.
     #[clap(required(true))]
     pub input_file: Vec<String>,
+    /// Combine all policies.
```
dburgener, line 19: This comment needs to specify that we build them with the name of the argument, something like "Combine all policies into one system and generate a cil file named OUT_FILENAME".

src/bin/casc/args.rs
```diff
@@ -1,13 +1,25 @@
...
 
 #[derive(Parser, Debug)]
-#[clap(author, version, name = "casc", about = "Compile Cascade SELinux policies into CIL", long_about = None)]
+#[clap(
+    author,
+    version,
+    name = "casc",
+    about = "Compile Cascade SELinux policies into CIL",
+    long_about = "Compile Cascade SELinux policies into CIL
+Supported use cases:
```
dburgener, line 10: Thanks for updating this way.  Upon seeing it, I think I'm reconsidering the list approach at all.

I think I'd rather push as much description as possible down to the OPTIONS section below.  Currently a fair amount is duplicated between DESCRIPTION and OPTIONS.

I think what might make more sense is a first paragraph saying "Compile..." just like now.  Then a second paragraph that explains that the -o and -s options are mutually exclusive and referring the user to the OPTIONS section.

src/bin/casc/args.rs
```diff
@@ -1,13 +1,24 @@
...
 
 #[derive(Parser, Debug)]
-#[clap(author, version, name = "casc", about = "Compile Cascade SELinux policies into CIL", long_about = None)]
+#[clap(
+    author,
+    version,
+    name = "casc",
+    about = "Compile Cascade SELinux policies into CIL",
+    long_about = "Compile Cascade SELinux policies into CIL.
+The -o option to combine all policies and the -s option to build individual systems are mutually exclusive. See the OPTIONS section for more information on these options."
```
dburgener, line 10: I think double newlines are needed to make this render as multiple paragraphs?

src/bin/casc/args.rs
```diff
@@ -1,13 +1,24 @@
...
+)]
 pub struct Args {
-    /// List of input files to process.  Directories are searched recursively.
+    /// List of input files to process. Directories are searched recursively.
     #[clap(required(true))]
     pub input_file: Vec<String>,
+    /// This is the default.
```
dburgener, line 16: Can we say "default behavior" instead of just "default"?

dburgener (approved):

PR #52, opened 2022-09-09
Fix handling of the "any" keyword in file contexts

dburgener:
The "any" keyword makes a file context match any object class. Previously, this was supposed to be indicated by the empty string, but that didn't actually make any sense, because it's in an unquoted context and whitespace is skipped by the lexer, so there's no actual way to indicate an empty string file context.

Fix this by using the keyword "any" (the new Cascade keyword now matches the CIL keyword with the same purpose).

Also, expand the file_context test to test the any keyword


matt-sheets (approved):
The only very nitpicky comment I have is to update the types in the test to have _t (foo -> foo_t) to coincide with standard ref policy naming conventions.  But since it just a test I am not heavily invested in it.

PR #54, opened 2022-09-23
Fix issues in automatic classperm generation

dburgener:
* Set of missed perms in process object class
* Typo in node object class


matt-sheets (approved):
Only comment is it might be a good time to add the rest of the process accesses.  The only ones I see missing are:
* execmem
* execstack
* execheap
* setkeycreate

dburgener:
Good catch, thanks.  Not sure why my system didn't complain about those as well.  Will add and update PR.

dburgener:
> Not sure why my system didn't complain about those as well.

Follow-up on this.  It did complain about those, I just didn't read the log closely enough.

PR #55, opened 2022-09-23
Fix new clippy warnings

dburgener:
Introduced by latest clippy version.

https://rust-lang.github.io/rust-clippy/master/index.html#manual_retain

Manual find
(https://rust-lang.github.io/rust-clippy/master/index.html#manual_find) provides a broken suggested fix and seems to require an extra unneeded copy to work around the fact that find adds a layer of indirection vs the implemented approach.

### DIFF
--- a/src/internal_rep.rs
+++ b/src/internal_rep.rs
@@ -371,14 +371,10 @@ impl TypeInfo {
     // It's possible to inherit from multiple built-ins, so order matters here.  We return the
     // first type in order of preference.
     pub fn get_built_in_variant(&self, types: &TypeMap) -> Option<&str> {
-        for t in constants::BUILT_IN_TYPES {
-            if self.is_type_by_name(types, t) {
-                return Some(t);
-            }
-        }
-
-        // I don't think this should be logically possible
-        None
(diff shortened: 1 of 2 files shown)


src/internal_rep.rs
```diff
@@ -370,6 +370,8 @@ impl TypeInfo {
     // All types must inherit from some built in.  Get one for this type.
     // It's possible to inherit from multiple built-ins, so order matters here.  We return the
     // first type in order of preference.
+    // Using find gets us a &&str, and dereferencing inserts an unnecessary copy
+    #[allow(clippy::manual_find)]
```
l0kod, line 374: `inter()` and `into_iter()` on a slice return the same thing, which is an issue to follow this tips. However, we can do `constants::BUILT_IN_TYPES.iter().find(|t| self.is_type_by_name(types, t)).copied()`, which will only copy a pointer/reference, so I think it does the same as the current code.

dburgener: Thanks for the suggestion.  Switching to that seems to work.

l0kod (approved):

PR #57, opened 2022-10-14
Improve module and system documentation

dburgener:
* Add syntax examples
* Remove mention that traits are future work (they have now been implemented)
* Split paragraphs
* Simplify wording around virtual modules

### DIFF
--- a/doc/policy_composer.md
+++ b/doc/policy_composer.md
@@ -1,16 +1,26 @@
 # Modules
-Modules are made up of domains, resources, traits (still in development) and
-other modules. They are the building blocks for larger scale policy and are a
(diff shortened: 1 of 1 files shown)


matt-sheets:
Do we want to wait to merge this guy until we decide what the new system keyword will be or push this guy forward and fix the system keyword later?

dburgener:
Good question.  I think that this has independent value, and it will be simple and straightforward to go through and change the keyword later, so we should go ahead with this and then update the docs in line with the keyword rename (assuming that is the solution we go with).

matt-sheets (approved):

PR #58, opened 2022-10-18
Detect whether we are writing to a terminal and conditionally enable color

dburgener:
Detect whether we are writing to a terminal and conditionally enable color on error output

I had assumed termcolor did this for us, but according to their documentation, they do not, and recommend using the atty crate for this purpose: https://docs.rs/termcolor/latest/termcolor/#detecting-presence-of-a-terminal

The approach in this commit lets casc determine whether or not to use color, and then error.rs translates that boolean decision into the appropriate termcolor commands.  Casc makes its decision in turn based on atty detecting a terminal.  This architecture allows for future binaries to make different decisions if needed (eg a --nocolor flag).

Reported-by: Matthew Sheets <masheets@linux.microsoft.com>

### DIFF
--- a/src/bin/casc/args.rs
+++ b/src/bin/casc/args.rs
@@ -20,8 +20,11 @@ pub struct Args {
     #[clap(default_value = "out.cil", short, value_parser = clap::builder::ValueParser::new(parse_out_filename))]
     pub out_filename: String,
     /// Build the systems from the SYSTEM_NAMES list. "-s all" to build all defined systems.
-    #[clap(short, multiple_values = true, conflicts_with = "out-filename")]
+    #[clap(short, conflicts_with = "out-filename")]
     pub system_names: Vec<String>,
+    ///colorize the output.  WHEN can be 'always', 'auto' (default), or 'never'
+    #[clap(long, value_enum, id = "WHEN")]
+    pub color: Option<ColorArg>,
 }
 
 fn parse_out_filename(filename: &str) -> Result<String, String> {
@@ -29,4 +32,11 @@ fn parse_out_filename(filename: &str) -> Result<String, String> {
         return Ok(filename.to_string());
     }
     Err(String::from("The value does not end in \".cil\""))
+}
+
+#[derive(clap::ValueEnum, Clone, Debug)]
+pub enum ColorArg {
+    Always,
+    Auto,
+    Never,
 }
--- a/src/bin/casc/main.rs
+++ b/src/bin/casc/main.rs
@@ -4,12 +4,13 @@ use selinux_cascade::error::{CascadeErrors, ErrorItem};
 use selinux_cascade::{compile_combined, compile_system_policies, compile_system_policies_all};
 
 mod args;
-use args::Args;
+use args::{Args, ColorArg};
 
 use clap::Parser;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::{Error, ErrorKind, Write};
+use termcolor::ColorChoice;
 use walkdir::WalkDir;
 
 fn main() -> std::io::Result<()> {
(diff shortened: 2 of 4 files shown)


src/bin/casc/main.rs
```diff
@@ -76,9 +76,9 @@ fn main() -> std::io::Result<()> {
 fn print_error(error_list: CascadeErrors) -> std::io::Result<()> {
     for e in error_list {
         if let ErrorItem::Parse(p) = e {
-            p.print_diagnostic();
+            p.print_diagnostic(atty::is(atty::Stream::Stderr));
```
l0kod, line 79: I'd like to be able to keep colors even when not using a terminal e.g., piping into `less -R`. I then suggest to add a casc option `--color=WHEN` that defaults to `auto` (cf. `ls`or `grep` options).

src/bin/casc/args.rs
```diff
@@ -20,8 +20,11 @@ pub struct Args {
     #[clap(default_value = "out.cil", short, value_parser = clap::builder::ValueParser::new(parse_out_filename))]
     pub out_filename: String,
     /// Build the systems from the SYSTEM_NAMES list. "-s all" to build all defined systems.
-    #[clap(short, multiple_values = true, conflicts_with = "out-filename")]
+    #[clap(short, conflicts_with = "out-filename")]
     pub system_names: Vec<String>,
+    ///colorize the output.  WHEN can be 'always', 'auto' (default), or 'never'
```
matt-sheets, line 25: When running casc --help the output for the color arg is indented further than the others.

matt-sheets:
Just for my own selfish learning are the `#[clap` & `#[derive` tags used for documentation like doxygen, or do they do more fancy things?

dburgener:
They do fancy things.  They are both [attribute like macros](https://doc.rust-lang.org/book/ch19-06-macros.html), which transform into some code which is then compiled.

#[derive()] is a very common macro in rust programming that lets you automatically implement a trait with automatic implementation of its functions.  You can read more about it [here](https://doc.rust-lang.org/book/appendix-03-derivable-traits.html) and [here](https://doc.rust-lang.org/book/ch10-02-traits.html).

As an example, this PR derives the Debug trait, which allows an enum to be printed in a programmer-understandable format in debug output.  In order to do that, all members of the struct or enum must also be Debug, and then rust can combine them all together to determine what a reasonable Debug output would be.  Of course, rust's decisions may not be the ones the programmer would have made in every case, in which case the programmer would need to implement the trait themselves with an impl {} block to define their own implementation.  You can see an example of that [here](https://github.com/dburgener/cascade/blob/6a9dff0a4124c8f39c84a50016cd6da043c55eb9/src/internal_rep.rs#L205).  The default derivation of PartialEq would return true if all fields are equal and false otherwise.  In our case, we want to say that two TypeInfos are equal if they share a name, regardless of the other fields.  Implementing it ourselves also means that the children don't all need the trait.  In this case, we would get a compiler error if we attempted to derive PartialEq, because in attempting to compare the equality of all fields, rust would see that BoundTypeInfo doesn't implement PartialEq, and therefore it doesn't know how to compare two BoundTypeInfos.

The #[clap()] attribute is defined by the [clap](https://docs.rs/clap/latest/clap/) library, which is a dependency.  Unfortunately, their documentation is all on the syntax for version 4 of the library and we're still on version 3, so the docs no longer match what we have here.  The clap library is evolving quickly and continues to change syntax aggressively.  But in short, they've defined macros to implement the command line parsing functionality based on our attributes and struct defintions.  The way that works in clap 4 is detailed in their "Derive Tutorial" at the above link.  It's highly similar to clap 3, but has some syntactic differences, including not actually having a "clap" attribute anymore.

Migrating to clap 4 is on my todo list, but isn't urgent.

src/bin/casc/args.rs
dburgener: Interesting, thanks for pointing this out.

It looks to me like what is happening is that clap is aligning short args and long args.  Since --color doesn't have a short form, it's aligned to the long form of the other args.

I agree that this looks a little weird, but it's in line with (eg) rustc, which I imagine is a comparison point for the clap developers.

In a quick googling, I don't see any way that we have control over this.  It seems to be a clap design choice in how they display the --help string.  Things may possibly change when we migrate to clap 4.

PR #61, opened 2022-10-27
Prevent obvious duplicate inherits

matt-sheets:
Previously we allow the user to do something like: resource baz inherits foo, foo, bar

This is either a needless duplicate or a possible typo. The compiler will now produce an error if a duplicate inherit like this is present.


dburgener:
You're failing the "cargo fmt" check.  You'll need to get the rustfmt package via rustup (https://github.com/rust-lang/rustfmt) and then run cargo fmt, which will automatically clean up the formatting.

The next check after that, which didn't run because fmt failed is clippy, so you may want to also get and run clippy as well (https://github.com/rust-lang/rust-clippy).  The exact clippy command used by the CI is available in .github/workflows/clippy.sh.  You'll want to make sure you use "-A clippy::new_without_default" to avoid unrelated failures.

src/internal_rep.rs
```diff
@@ -248,6 +248,24 @@ impl Declared for TypeInfo {
 
 impl TypeInfo {
     pub fn new(td: TypeDecl, file: &SimpleFile<String, String>) -> Result<TypeInfo, CascadeErrors> {
+        let mut temp_vec = td.inherits.to_vec();
+        temp_vec.sort();
+        let mut iter = temp_vec.iter().peekable();
+        while let Some(cur_val) = iter.next() {
```
dburgener, line 254: I think the `let` followed by `while let` pattern here should be able to be written more cleanly as `for cur_val in temp_vec.iter().peekable() {`?

src/internal_rep.rs
```diff
@@ -248,6 +248,24 @@ impl Declared for TypeInfo {
...
+            match iter.peek() {
```
dburgener, line 255: A match that only has one active branch is better as an `if let`.  (`if let Some(next_val) = iter.peek() { ... }`)

dburgener: (I suspect if you run clippy on this it will make this suggestion with the exact code to use).

src/internal_rep.rs
```diff
@@ -248,6 +248,24 @@ impl Declared for TypeInfo {
...
+                Some(next_val) => {
+                    if cur_val == *next_val {
+                        return Err(CascadeErrors::from(ErrorItem::Compile(CompileError::new(
+                            "Duplicate Inherit",
+                            file,
+                            (*next_val).get_range(),
+                            "Duplicate Inherit. Possible Typo?",
```
dburgener, line 262: The additional help here shouldn't restate what the first message said directly.  I'd prefer something more verbose, like "This type to inherit is identical to another type in the same inheritance list.  Perhaps you meant to inherit some other type?"

src/lib.rs
```diff
@@ -1090,4 +1090,18 @@ mod tests {
...
     }
+
+    #[test]
+    fn invalid_duplicate_inherit() {
+        error_policy_test!("duplicate_inherit.cas", 2, ErrorItem::Compile(_));
+    }
+
+    #[test]
+    fn valid_duplicate_inherit() {
+        valid_policy_test("duplicate_inherit.cas", &[
```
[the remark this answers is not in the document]
matt-sheets: I built that cas file with what is on main and "typeattributeset foo (qux)" is missing.  I agree this might be a short coming with the compiler.  But I haven't mucked around with the cil stuff enough to say if having "(typeattributeset foo (baz))" & "(typeattributeset baz (qux))" is good enough.

src/internal_rep.rs
matt-sheets: We need the iterator object to be able to call peek on it.  Because of that I don't think we can do a for in loop here.

dburgener: Yeah, I guess you're right about that.  That's too bad.

src/lib.rs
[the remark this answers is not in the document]
dburgener: Okay, I read the docs and poked at secilc a bit, and you are absolutely right.  CIL handles this for us via transitive typeattribueset.  So this is a non-issue.

src/internal_rep.rs
```diff
@@ -248,6 +248,21 @@ impl Declared for TypeInfo {
 
 impl TypeInfo {
     pub fn new(td: TypeDecl, file: &SimpleFile<String, String>) -> Result<TypeInfo, CascadeErrors> {
+        let mut temp_vec = td.inherits.to_vec();
```
dburgener, line 251: td.inherits is already a vector.  I think just `clone()` instead of `to_vec()` would be clearer?

dburgener (approved):

dburgener:
I tried to squash and merge this using the github interface, but I'm running in to this issue: https://github.com/isaacs/github/issues/1368

Instead of saying the commit was authored by masheets@linux.microsoft.com as desired, it's using an internally generated users.noreply.github.com e-mail address for you, that I assume is set as a default in your settings somewhere (probably the "Keep my e-mail address private" box at https://github.com/settings/emails).

Are you able to uncheck the box above so I can hopefully get this to squash merge with the right attribution?

matt-sheets:
> Are you able to uncheck the box above so I can hopefully get this to squash merge with the right attribution?

I updated that setting, I would try again.

dburgener:
Merged.  Github still messed it up a little bit as your author name in the original commits was "Matt Sheets", but the github squash merge changed it to "matt-sheets" without warning me that it would...  But at least the e-mail address is correct.  That might be because your "Name" https://github.com/settings/profile is set to "matt-sheets"?

The whole thing is kind of obnoxious.  I may just stick to manual squash and force push in the future.

Regardless, this is now merged.  Thanks for the contribution!

PR #62, opened 2022-10-28
Add dbus_contexts template

dburgener:
Some info about dbus and SELinux: https://blog.siphos.be/2014/06/d-bus-and-selinux/

SELinux policies ship a dbus_contexts file at
/etc/selinux/[policyname]/contexts/dbus_contexts which provides mappings from dbus service names to SELinux label.  If a mapping is not defined, dbus uses the label of the owning process.

The common case is to leave this file empty as with this template and simply allow the default behavior.

Dbus throws errors and has substantially degraded functionality if this file is not present at all, so we need to ship a file there.

The current approach doesn't yet provide support for actually specifying the mappings, but it is full featured enough to make that straightforward when it is added.  Presumably something like an @dbus() annotation will be a very straightforward way to specify this in Cascade, and it can then be plugged in to this machinery.


dburgener:
Throwing this quick and dirty early code up as a draft/RFC in case anyone has feedback.  Major TODOs that I intend to implement before this is ready for merge include:

* Hooking into the library API and casc
* Error handling

But the core functionality is there for now, and the overall approach/dependency might be worth commenting on.

dburgener:
I made a PR for the --package frontend here: https://github.com/dburgener/cascade/pull/71

The idea is that once that merges, this can get integrated with that and the dbus_contexts file can be included in the package.

dburgener:
Pushed an update integrating this with the --package flag.  It still needs error cleanup before it's ready for review.

dburgener:
All of the above mentioned issues have now been addressed, so I'm removing the RFC tag and marking this as ready for review.

matt-sheets:
Do we want to add a new test to generate the dbus file and compare it to a known good like we do with the cil?

matt-sheets:
> > Do we want to add a new test to generate the dbus file and compare it to a known good like we do with the cil?
> 
> I'm not sure I see the value add past the test already in dbus.rs? The test in dbus.rs is verifying that the full string is equal to a known value, which is effectively the same as the cil tests do. The only difference is not fully writing out the file, but I'm not sure that's something I see high value in testing.

I was assuming at some point the debus context may be populated with more meaningful data and having the infrastructure in place would make that easier to test going forward.  But we can cross that bridge if the time ever comes.

dburgener:
I hear you, but I think we should generally be able to add those tests in dbus.rs where we already have the basics in place.  Even once we expand, I don't think these tests need to be particularly more heavy-weight.

I imagine/hope that most of the real logic will stay in dbus.rs, and the lib.rs wrapper will remain very thin.

matt-sheets (approved):

PR #67, opened 2022-11-11
Implement Type Casting

dburgener:


matt-sheets:
You should add some negative tests cases as well as the sunny day one you have currently

matt-sheets (approved):
I have some educational questions about the cow borrow thing, but I will ping you about that offline.

PR #70, opened 2022-11-17
Split collapsed object classes in outputted CIL.

dburgener:
The capability, process and cap_userns object classes have more than 32 permissions, which is the max permissions per class.  Therefore SELinux has separate capability2, process2 and cap2_userns classes to handle this overflow.  In Cascade, we just treat these "2" permissions as part of the main class.

Previously, validation worked, but the outputted CIL didn't handle the mapping correctly.  This commit correctly maps capability2, process2 and cap2_userns permissions back to the correct object class in the output.


matt-sheets:
Will I be able to use both capability and capability2 in cascade or just capability?

dburgener:
In the long term, no, the plan is to only allow you to use capability.  However, currently only converting capability to capaibility2 is implemented, and using capability2 directly still works.

matt-sheets (approved):

PR #71, opened 2022-11-21
Add --package argument to casc

dburgener:
--package takes the output files and packages them up into the contents to go in /etc/selinux

Currently, we only package the compiled policy.32 and file_contexts. This commit assumes that the secilc version present generates a policy.32.  That's an obviously broken assumption that should be fixed before merging this.  Over the longer term, we should be able to take the infrastructure added here and add all the other files.

For a policy built without specific system configuration, this names the final policy "out", which may not be desirable.


dburgener:
Marking as draft for now, because of two issues described in the commit message.  I think they should probably both be addressed before merging this, but I wanted to get it up for review and discussion in the interim.  Particularly to solicit suggestions on the issue around policy naming if the user hasn't specified any name.

dburgener:
I've marked this as ready for review.  I've added a new commit that hardcodes the use of policy.32 for the time being.  That's better than the initial version which assumed secilc would output a policy.32.  In the original version, packaging would be broken on a modern secilc that outputs policy.33.  With this update, that secilc will simply output a policy.32 and packaging will work.

Obviously, we'll want to use policy.33 on modern secilc, but that's a separate task that IMO shouldn't block this.  This sets up the infrastructure for configurable secilc versions, and will be easily pluggable once we set up secilc version handling (which we also need for #75)

On the topic of naming the final policy "out", Mickael pointed out in an offline discussion that we can override the policy name both with a machine {} block or the -o command line flag.  With multiple straightforward workarounds, I'm not sure the default matters so much that it should block this PR.  I'm open to pushback on that if others disagree.

matt-sheets (approved):

PR #74, opened 2022-11-23
Rename "system" to "machine"

### LINKED ISSUE #59: `system` is both a Cascade keyword and an object class
A rule like:

```
allow(this, foo, system, [start]);
```

Fails to compile currently, because `system` is parsed as the reserved keyword rather than some arbitrary string that will later be consumed as a valid object class.

dburgener:
The word "system" already refers to the SELinux object class "system". Rather than adding complexity differentiating the two uses of the term, just rename our term to "machine" to avoid the name conflict.

closes #59


matt-sheets (approved):

src/bin/casc/args.rs
```diff
@@ -8,20 +8,20 @@ use clap::Parser;
...
     #[clap(required(true))]
     pub input_file: Vec<String>,
     /// This is the default behavior.
-    /// Combine all policies into a monolithic system policy with system configuration options set to default values.
+    /// Combine all policies into a monolithic machine policy with machine configuration options set to default values.
     /// The generated CIL file is named OUT_FILENAME.
     #[clap(default_value = "out.cil", short, value_parser = clap::builder::ValueParser::new(parse_out_filename))]
     pub out_filename: String,
-    /// Build the systems from the SYSTEM_NAMES list. "-s all" to build all defined systems.
+    /// Build the machines from the MACHINE_NAMES list. "-s all" to build all defined machines.
```
dburgener, line 22: The short version of this will now be "-m" I imagine.  We should verify that and change the "-s" direction.

PR #75, opened 2022-11-28
Adding support for the fs_context function

matt-sheets:
The user can now use "fs_context" to implement fsuse and genfscon rules in policy.


src/internal_rep.rs
```diff
@@ -1329,6 +1339,257 @@ fn call_to_fc_rules<'a>(
...
+        .next()
+        .ok_or_else(|| ErrorItem::Internal(InternalError::new()))?
+        .get_name_or_string(context)?;
+    let fscontext_type = match fscontext_str.to_string().parse::<FSContextType>() {
+        Ok(f) => f,
+        Err(_) => {
+            return Err(CascadeErrors::from(ErrorItem::Compile(CompileError::new(
+                "Not a valid file system type",
+                file,
+                fscontext_str.get_range(), //TODO error not showing correctly
```
dburgener, line 1530: Is this TODO still an issue?

src/internal_rep.rs
```diff
@@ -1329,6 +1339,257 @@ fn call_to_fc_rules<'a>(
...
+                "File system type must be 'xattr', 'task', 'trans', or 'genfscon'",
+            ))));
+        }
+    };
+    let regex_string = args_iter
```
dburgener, line 1535: We want to error if someone has provided an FSContextType other than genfscon and also provided a path or file_type, right?  I don't see that here.  Am I missing it?

src/internal_rep.rs
```diff
@@ -1814,6 +2077,21 @@ impl<'a> ValidatedStatement<'a> {
...
                 }
+                Some(BuiltIns::FileSystemContext) => {
+                    if in_resource {
+                        Ok(call_to_fsc_rules(c, types, class_perms, &*context, file)?
+                            .into_iter()
+                            .map(ValidatedStatement::FscRule)
+                            .collect())
+                    } else {
+                        Err(CascadeErrors::from(ErrorItem::Compile(CompileError::new(
+                            "file_context() calls are only allowed in resources",
```
dburgener, line 2088: copy/paste update error.

data/policies/fs_context.cas
```diff
@@ -0,0 +1,20 @@
+resource foo {
```
dburgener, line 1: All the policy test files added in this PR are mixing spaces and tabs.

dburgener (changes requested):

doc/TE.md
```diff
@@ -278,6 +283,38 @@ TODO: default contexts.  This will be part of the next phase of design
...
+`fs_context(resource fs_label, string fs_name, fs_type type, path path, file_type [obj_class])`
+
+* `fs_label` is label you wish to apply to the filesystem.
+* `fs_name` is the OS recognized name for the filesystem. (e.g. ext4, proc, tmpfs)
+* `fs_type` must be one of the following options:
+  * `xattr`: Filesystems supporting the extended attribute security.selinux. The labeling is persistent for filesystems that support extended attributes.
+  * `task`: For pseudo filesystems supporting task related services such as pipes and sockets.
+  * `trans`: For pseudo filesystems such as pseudo terminals and temporary objects.
+  * `genfscon`: Used to allocate a security context to filesystems that cannot support any of the previous file labeling options
+* `path` is an optional path relative to root of the mounted filesystem.  Currently this is only valid for the proc filesystem, all other types must be "/".  If not given, the field will default to "/".
```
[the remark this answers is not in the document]
matt-sheets: Just going from the cil documentation: https://github.com/SELinuxProject/selinux/blob/master/secilc/docs/cil_file_labeling_statements.md#genfscon

src/internal_rep.rs
matt-sheets: no deleting

doc/TE.md
```diff
@@ -278,6 +283,38 @@ TODO: default contexts.  This will be part of the next phase of design
 
 TODO: file_contexts.subs_dist (although is it necessary with advanced file_context handling features?)
 
+### Filesystem labeling
```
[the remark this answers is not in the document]
matt-sheets: I used the rewrap extension in vscode.  I wouldn't be surprised if there is some awk command that will do something similar.

src/internal_rep.rs
```diff
@@ -1338,6 +1348,277 @@ fn call_to_fc_rules<'a>(
...
+        .ok_or_else(|| ErrorItem::Internal(InternalError::new()))?;
+    let regex_string = regex_string_arg.get_name_or_string(context)?.to_string();
+    let file_types_arg = args_iter
+        .next()
+        .ok_or_else(|| ErrorItem::Internal(InternalError::new()))?;
+    let file_types = file_types_arg.get_list(context)?;
+
+    match fscontext_type {
+        FSContextType::XAttr | FSContextType::Task | FSContextType::Trans => {
+            // The 'regex_string_arg.get_range().is_none()' is a hacky way to
```
dburgener, line 1556: Eww.  I'm not immediately sure what a better answer is though.  This is mostly a note to self to think about this more.  I'm worried that overloading the usage of the range like this will cause us problems down the road, since we're tying this user to the implementation detail.  I can't think of any reason it wouldn't work in the current code base though...

src/internal_rep.rs
```diff
@@ -1338,6 +1348,277 @@ fn call_to_fc_rules<'a>(
...
+            // to test if the actual arg was passed or not
+            if regex_string_arg.get_range().is_none() && file_types.is_empty() {
+                ret.push(FileSystemContextRule {
+                    fscontext_type,
+                    fs_name,
+                    path: None,
+                    file_type: None,
+                    context: fs_context.clone(),
+                });
+            } else if !file_types.is_empty() {
```
dburgener, line 1568: I think that if a user has made both mistakes we should return both errors at once.

src/error.rs
```diff
@@ -214,6 +214,21 @@ impl InvalidSystemError {
     }
 }
 
+#[derive(Error, Clone, Debug)]
+#[error("{diagnostic}")]
+pub struct InvalidFileSystemError {
+    pub diagnostic: Diag<usize>,
+}
+
+impl InvalidFileSystemError {
```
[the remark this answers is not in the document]
matt-sheets: I used this new error type in places that I didn't know how to get the file from a given method or may be caused by fs_context declarations from multiple different places.

I just may be doing something wrong though.

src/compile.rs
```diff
@@ -312,6 +326,81 @@ pub fn build_func_map<'a>(
...
+        for rule in &v {
+            match rule.fscontext_type {
+                FSContextType::XAttr | FSContextType::Task | FSContextType::Trans => {
+                    // If we ever see a duplicate and one of them is xattr, task, or trans something is wrong
+                    // return an error and look for more.
+                    errors.append(CascadeErrors::from(InvalidFileSystemError::new(&format!(
+                        "Duplicate filesystem context.\n Found two different filesystem type declarations for filesystem: {}",
+                        rule.fs_name
+                    ))));
+                    continue 'key_loop;
```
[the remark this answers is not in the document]
matt-sheets: You would see duplicate errors, one in each way if that makes sense.

Let's say we have two fs_context for /zap/ one with label bar and one with label foo we would see:

``` 
Found duplicate genfscon rules with differing contexts: "test"
        Path: "/zap/baa"
        Context 1: system_u:object_r:bar:s0 - s0
        Context 2: system_u:object_r:foo:s0 - s0
```
And we would also see
```
 Found duplicate genfscon rules with differing contexts: "test"
        Path: "/zap/baa"
        Context 1: system_u:object_r:foo:s0 - s0
        Context 2: system_u:object_r:bar:s0 - s0
```

src/compile.rs
```diff
@@ -312,6 +326,81 @@ pub fn build_func_map<'a>(
...
+        if v.len() <= 1 {
+            continue;
+        }
+        for rule in &v {
+            match rule.fscontext_type {
+                FSContextType::XAttr | FSContextType::Task | FSContextType::Trans => {
+                    // If we ever see a duplicate and one of them is xattr, task, or trans something is wrong
+                    // return an error and look for more.
+                    errors.append(CascadeErrors::from(InvalidFileSystemError::new(&format!(
+                        "Duplicate filesystem context.\n Found two different filesystem type declarations for filesystem: {}",
```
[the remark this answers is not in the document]
dburgener: I think we'll need to save the file with the FileSystemContextRule in call_to_fsc_rules() so that we can use it here.  We have the file at that point, so it should just be a matter of storing it in the struct.

doc/TE.md
```diff
@@ -278,6 +283,38 @@ TODO: default contexts.  This will be part of the next phase of design
...
+  * `task`: For pseudo filesystems supporting task related services such as pipes and sockets.
+  * `trans`: For pseudo filesystems such as pseudo terminals and temporary objects.
+  * `genfscon`: Used to allocate a security context to filesystems that cannot support any of the previous file labeling options
+* `path` is an optional path relative to root of the mounted filesystem.  Currently this is only valid for the proc filesystem, all other types must be "/".  If not given, the field will default to "/".
+* `file_type` is an optional keyword representing a file type to apply the label to. Valid values are the same as in file_context function.  If not given, [any] is assumed.
+  * Note: You must use SELinux userspace tools version 3.4 or newer to use this field.
+
+`xattr`, `task`, and `trans` all represent filesystems that support SELinux security contexts.  The filesystem itself has a labeled applied to it as a whole, which is the `fs_label` provided in this function.  All files on the filesystem also store their own SELinux security context in their own extended attributes.
```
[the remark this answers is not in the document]
dburgener: Hmm, I think the rewording may have added to the confusion.  How about just enumerating the three cases for xattr, task and trans in terms of how overall filesystem relating relates to the labeling of files?

src/compile.rs
```diff
@@ -329,6 +344,84 @@ pub fn build_func_map<'a>(
     Ok(decl_map)
 }
 
+pub fn validate_fs_context_duplicates(
+    statements: &BTreeSet<ValidatedStatement>,
+) -> Result<(), CascadeErrors> {
+    let mut errors = CascadeErrors::new();
+    let mut fsc_rules: BTreeMap<&String, BTreeSet<&FileSystemContextRule>> = BTreeMap::new();
+
+    for statement in statements {
```
dburgener, line 353: I feel like this for loop could live at the validate_rules().  With it in here, we'll need to loop over the entire AST for each validation check we eventually add.  If we move it to the validate_rules() level and then pass fsc_rules to validate_fs_context_duplicates, then we can set up the data everyone needs in a single AST pass at the top level of validate_rules().

src/error.rs
```diff
@@ -80,15 +80,18 @@ impl CompileError {
     pub fn add_additional_message(
         mut self,
         file: &SimpleFile<String, String>,
-        range: Range<usize>,
+        range: Option<Range<usize>>,
```
dburgener, line 83: I'm currently in process on a commit to try to get rid of passing `Option<Range>`s  into errors.  It's certainly non-trivial, but the idea in my in progress commit is to force the caller to handle the None case in a more sane way, since the error layer doesn't have the context to understand what it means when the Range is none.

I haven't really looked much at the callers, but is the None case something that can be handled at the call site?  Perhaps by pointing at the thing that made a synthetic type with a None Range?

src/internal_rep.rs
```diff
@@ -1539,6 +1552,316 @@ fn call_to_fc_rules<'a>(
...
+                    // We are purposefully falling through without an else to
+                    // reduce redundant lines of code
+                    Ok(list(&[
+                        atom_s("genfscon"),
+                        atom_s(f.fs_name.trim_matches('"')),
+                        atom_s(p.as_ref()),
+                        Sexp::from(&f.context),
+                    ]))
+                } else {
+                    // We should never get here since we are defaulting to "/"
```
dburgener, line 1662: If we should never hit this in normal operation, should it be an InternalError instead?

src/compile.rs
```diff
@@ -329,6 +345,158 @@ pub fn build_func_map<'a>(
     Ok(decl_map)
 }
 
+pub fn validate_fs_context_duplicates(
```
dburgener, line 348: This function is long and deeply indented and therefore hard to follow.  Would it be possible to genericize the error cases a bit into a function or macro to make it more readable?  I see three instances of the "if let Some(unwrapped_error) = error" block with various differences that could perhaps be abstracted to make the whole thing more concise and readable.  I haven't read it all that closely yet though, so perhaps there's some challenges to that that make it not worth it?

dburgener (commented):
I didn't do a totally comprehensive review.  Just a few comments.

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
...
+                                    // error is not None so we have already found something, so we just
+                                    // need to add a new error message
+                                    if let Some(unwrapped_error) = error {
+                                        error = Some(unwrapped_error.add_additional_message(&inner_rule.file,
+                                            inner_range,
+                                            &format!("Found duplicate genfscon rules for filesystem {} with differing file types", inner_rule.fs_name)));
+                                    } else {
+                                        // Just like above we need to look through the outer function's args to get the correct range
+                                        let mut outer_range: Option<Range<usize>> = None;
+                                        for arg in &rule.func_call.args {
```
dburgener, line 440: This for loop can be its own function that takes the range and the rule, right?

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
...
+                                    } else {
+                                        let outer_range: Option<Range<usize>> =
+                                            rule.func_call.args[0].0.get_range();
+                                        // error is none so we need to make a new one
+                                        error = Some(CompileError::new(
+                                            "Duplicate genfscon contexts",
+                                            &rule.file,
+                                            outer_range,
+                                            &format!("Found duplicate genfscon rules for filesystem {} with differing contexts: {}", rule.fs_name, rule.context)));
+                                        // We just made the error we are okay to unwrap it
```
dburgener, line 407: Might be cleaner to just do `let tmp_error = CompilerError::new(...);` followed by `error = Some(tmp_error.add_additional_message(...)` in order to avoid creating an option just to immediately discard it?

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
     Ok(decl_map)
 }
 
+pub fn validate_fs_context_duplicates(
+    fsc_rules: BTreeMap<&String, BTreeSet<&FileSystemContextRule>>,
+) -> Result<(), CascadeErrors> {
+    let mut errors = CascadeErrors::new();
+
+    'key_loop: for (_, v) in fsc_rules {
```
dburgener, line 355: I think this is equivalent to `for v in fsc_rules.values()` and the values() version is more idiomatic.

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
...
+                FSContextType::GenFSCon => {
+                    // genfscon gets more complicated.  We can have similar rules as long as the paths are different.
+                    // If we find a genfscon with the same path, they must have the same context and object type.
+                    if let Some(path) = &rule.path {
+                        // Look through the rules again
+                        for inner_rule in &v {
+                            // Only check path if it was provided as part of the rule
+                            if let Some(inner_path) = &inner_rule.path {
+                                // If our paths match, check if our contexts match
+                                if path == inner_path && rule.context != inner_rule.context {
```
dburgener, line 388: The implication here is that duplicate genfscon rules are acceptable if the contexts are identical?

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
...
+                                    // In this case we are safe in knowing args[0] is what we need.
```
dburgener, line 389: Is this actually true?  With named arguments, I think the args could come in any order.

src/compile.rs
dburgener: Actually, if we store the range in FileSystemContextType, we don't need this loop at all I think.

src/internal_rep.rs
```diff
@@ -1539,6 +1552,311 @@ fn call_to_fc_rules<'a>(
...
+            "trans" => Ok(FSContextType::Trans),
+            "genfscon" => Ok(FSContextType::GenFSCon),
+            _ => Err(()),
+        }
+    }
+}
+
+#[derive(Derivative)]
+#[derivative(Clone, Debug, PartialEq, PartialOrd, Ord)]
+pub struct FileSystemContextRule<'a> {
```
dburgener, line 1593: We should generally be storing CascadeStrings here so that we don't need to go back to the FuncCall later to get ranges.

src/internal_rep.rs
```diff
@@ -1539,6 +1552,311 @@ fn call_to_fc_rules<'a>(
     Ok(ret)
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum FSContextType {
```
dburgener, line 1556: It looks like we want the range of this later.  So the top level type should probably be a struct that holds this enum, and a range.

src/compile.rs
matt-sheets: Yes, my thought was that it is benign and doesn't directly add ambiguity to the final policy so let the user do it.

src/compile.rs
matt-sheets: OBE

src/compile.rs
```diff
@@ -329,6 +347,158 @@ pub fn build_func_map<'a>(
...
+                                    let inner_range: Option<Range<usize>> =
+                                        inner_rule.func_call.args[0].0.get_range();
```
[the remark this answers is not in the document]
matt-sheets: OBE

src/compile.rs
```diff
@@ -329,6 +346,118 @@ pub fn build_func_map<'a>(
...
+                        // Look through the rules again
+                        for inner_rule in v {
+                            // Only check path if it was provided as part of the rule
+                            if let Some(inner_path) = &inner_rule.path {
+                                // If our paths match, check if our contexts match
+                                if path == inner_path && rule.context != inner_rule.context {
+                                    // error is not None so we have already found something, so we just
+                                    // need to add a new error message
+                                    if let Some(unwrapped_error) = error {
+                                        error = Some(unwrapped_error.add_additional_message(&inner_rule.file,
```
dburgener, line 383: This block looks identical to the similar one below except for the range.  I feel like defining a local closure that takes a range and then calling it here will be cleaner.

dburgener: Alternatively, why can't we just reuse add_or_create_compile_error() again?

src/internal_rep.rs
```diff
@@ -1545,6 +1565,370 @@ fn call_to_fc_rules<'a>(
...
+                    fscontext_type,
+                    fs_name,
+                    path: Some(regex_string),
+                    file_type: None,
+                    context: fs_context.clone(),
+                    file: file.clone(),
+                    // file_type_range shouldn't need to be used here since file_type is None, but I would rather not
+                    // have to deal with Option stuff later
+                    file_type_range: c.get_name_range().unwrap_or_default(),
+                    context_range: unwrap_or_return!(
```
dburgener, line 1888: I think it would be more idiomatic to convert the Option into a Result using [ok_or_else](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else) and then use the ? operator.

`context_str_arg.get_range().ok_or_else(|| CascadeErrors::from(InternalError::new())?`

src/internal_rep.rs
```diff
@@ -1545,6 +1565,370 @@ fn call_to_fc_rules<'a>(
...
+                        context_str_arg.get_range(),
+                        Err(CascadeErrors::from(InternalError::new()))
+                    ),
+                });
+            } else {
+                for file_type in file_types {
+                    let file_type = match file_type.to_string().parse::<FileType>() {
```
dburgener, line 1895: Third use case for https://github.com/dburgener/cascade/issues/89

I started implementing it, but the mildly annoying bit is that it needs the `file`, so just `TryFrom<&CascadeString>` like I previously said doesn't work.  So maybe it becomes just a regular function instead of a try_from(), idk.

src/internal_rep.rs
dburgener: Probably the long term answer will be to augment the TypeInstance with information about where this instance came from.  That seems like a longer term answer.  I think this way is fine for now.

src/compile.rs
matt-sheets: In the else we have a weird condition where we need to create the error then add a new message to it immediately.

src/internal_rep.rs
matt-sheets: I think this is obsoleted; I am able to get the range I need from the rule.

src/internal_rep.rs
```diff
@@ -3568,9 +3963,16 @@ mod tests {
             let mut renames = BTreeMap::new();
             renames.insert("old_name".to_string(), "new_name".to_string());
             let renamed_statement = statement.get_renamed_statement(&renames);
-            let sexp = Sexp::from(&renamed_statement);
-            assert!(sexp.to_string().contains("new_name"));
-            assert!(!sexp.to_string().contains("old_name"));
+            match Sexp::try_from(&renamed_statement) {
```
dburgener, line 3966: It's cleaner to just do "Sexp::try_from(&renamed_statement).unwrap()".  Unwrap is just panic on Err, so it's functionally equivalent but more concise and readable.

The general principle is that we don't panic in real code, and therefore don't use unwrap because it might panic.  But in test code, panicking is just test failure, so unwrap is idiomatic there.

src/compile.rs
```diff
@@ -329,6 +346,128 @@ pub fn build_func_map<'a>(
...
+    error: Option<CompileError>,
+    msg: &str,
+    file_a: &SimpleFile<String, String>,
+    file_b: &SimpleFile<String, String>,
+    range_a: Range<usize>,
+    range_b: Range<usize>,
+    help_a: &str,
+    help_b: &str,
+) -> CompileError {
+    let mut tmp_error;
```
dburgener, line 363: Now that this is in its own function (which is a huge improvement, IMO.  Thanks), this variable is probably better named "ret".

PR #78, opened 2022-12-01
Allow use of "self" keyword

matt-sheets:


src/lib.rs
```diff
@@ -1137,4 +1137,14 @@ mod tests {
             &[],
         );
     }
+
+    #[test]
```
dburgener, line 1141: Can we also add a test for using "self" in the subject of an allow rule?  It should already be handled by the validate_arguments() code, but it would be nice to have some additional verification.

dburgener:
Thanks for implementing this!

One problem that's occurred to me, starting to look at the code is if we define a function that takes a resource as an argument, I think the current approach would allow passing "self" to it.  That seems weird, since "self" has a meaning dependent on local context.

We could check for this when we validate function calls, and maybe that's the right thing, but I'd love to do this in a way that most minimizes one-off checks.

matt-sheets:
I started implementing the check to not allow self to be passed in as a resource argument.  But I started thinking that passing self into a bunch of other stuff makes little to no sense either, things like file_context and domtrans.

I looked through reference policy v2 and it looks like the only place self is used is in dontaudit and allow statements.  Do we want to enforce the same with cascade?

src/compile.rs
```diff
@@ -201,6 +201,11 @@ pub fn get_built_in_types_map() -> Result<TypeMap, CascadeErrors> {
         built_in_types.insert(sid.name.to_string(), sid)?;
     }
 
+    // Add self as children of resource
```
dburgener, line 204: nit: should be "child" singular

src/internal_rep.rs
```diff
@@ -2578,6 +2585,23 @@ fn validate_argument<'a>(
...
+                        .param_type
+                        .is_child_or_actual_type(resource_type, types))
+                    && arg_typeinfo.name == constants::SELF
+                    && !is_avc
+                {
+                    return Err(ErrorItem::Compile(CompileError::new(
+                        "'self' passed as resource argument",
+                        file,
+                        arg.get_range(),
+                        "'self' cannot be passed into a function as a resource since 'self' is dependent on local context",
```
dburgener, line 2600: Thanks, this is nice.

Looking at the test case, I could imagine someone making this mistake because they wanted to use "this" instead.  What do you think about adding a parenthetical note that says "(If you intended to refer to the type declared in the enclosing block, use "this")"?

dburgener:
Merged, thanks!

PR #79, opened 2022-12-02
Build multiple secilc versions in our github actions

dburgener:


dburgener:
This doesn't do anything yet about supporting variable cascade behavior based on secilc version, but it does test against multiple versions, so we can 1. ensure we remain regression free across multiple versions and 2. Have testing infrastructure ready to go once we support variable behavior based on target secilc version.

The current testing only goes back to 3.0.  Ideally I'd like it to go back to 2.7.  See the third commit message for details.

The plan is to first merge this and get the improved test coverage.  Then we can add support for multiple secilc versions into cascade and pass the version information in an environmental variable from the github job to cargo test.  Then cargo test can build policies using the appropriate secilc version flags.  For if no environmental variable is supplied, cargo test will use a sensible default output target (3.0 seems like a good default).  Since cil features are generally backwards compatible, targeting 3.0 should work across all versions > 3.0 as well.

matt-sheets (approved):

PR #81, opened 2022-12-12
Add support for X server object classes

dburgener:


src/obj_class.rs
```diff
@@ -109,6 +109,26 @@ const COMMON_CAP2_PERMS: &[&str] = &[
...
+    "bell",
+    "force_cursor",
+    "freeze",
+    "grab",
+    "manage",
+    "list_property",
+    "get_property",
+    "set_property",
+    "add",
+    "remove",
```
matt-sheets, line 129: In the newest ref policy, there is also [create and destroy](https://github.com/SELinuxProject/refpolicy/blob/master/policy/flask/access_vectors#L126). This may affect x_pointer and x_keyboard below.

dburgener: Interesting.  I was going based off of https://selinuxproject.org/page/ObjectClassesPerms#common_x_device

I dug into this a bit, and I found the x windows security classes in the x server code here: https://github.com/freedesktop/xorg-xserver/blob/722da1c62caad961ce09fc226daf84465f4b7a7e/Xext/xselinuxint.h

What's interesting is that they don't include x_device at all.  It looks like it was actually replaced with x_pointer and x_keyboard in 2009: https://github.com/freedesktop/xorg-xserver/commit/6c2ae5fec552366e11ad64a27626eb5dec4becf0

So presumably the x_device permissions in refpolicy are to support legacy versions of x windows.

The 2009 version of x_device did include "create" and "destroy", so I guess selinuxproject.org is just wrong on that topic.

I think a question worth asking is whether we want our defaults to support 2009 era x windows at all (we will eventually support custom object class definitions).

My thought is that it might make sense to just eliminate the x_device object class for Cascade auto-generated classperms, and in the unlikely event that someone in the future is trying to use pre-2009 x windows with a Cascade policy, they will be able to custom define the object class.

Thoughts?

matt-sheets: I think that's good.  Once refpolicy3 is a usable thing I doubt folks will try to get it running on 10+ year old xwindows.

PR #82, opened 2022-12-12
Add support for parent call function syntax

matt-sheets:
Users will now be able to use the foo\<bar\>.read() syntax


matt-sheets:
As of posting this draft, we are doing this the "old" way of <parent> instead of ::parent.  But the rest of the logic should be sound.

src/internal_rep.rs
```diff
@@ -2320,11 +2320,36 @@ impl ValidatedCall {
...
 
-        // Each argument must match the type the function signature expects
-        let mut args = match &call.class_name {
-            Some(c) => vec![convert_class_name_if_this(c, parent_type)?.get_cil_name()],
-            None => Vec::new(),
-        };
+        let mut args =
+            match &call.parent_name {
+                Some(parent_name) => {
+                    match function_info.class {
```
dburgener, line 2326: This block is pretty hard to follow with all the indentation.  It looks like this whole match block could get pulled out into its own function (returning `Result<(), CascadeErrors>`), which would help readability a lot

src/internal_rep.rs
```diff
@@ -2320,11 +2320,36 @@ impl ValidatedCall {
...
+                        None => {
+                            return Err(CascadeErrors::from(ErrorItem::Compile(
+                                CompileError::new("No such class", file, call.get_name_range(), ""),
+                            )));
+                        }
+                    };
+                    vec![convert_class_name_if_this(parent_name, parent_type)?.get_cil_name()]
+                }
+                None => {
+                    // Each argument must match the type the function signature expects
```
dburgener, line 2346: I'm not really sure what the history of this comment is, but it sure seems irrelevant buried down in this block.  I'm not sure it's much better moved outside the block.  It can probably just be deleted, while you're rearranging things here.

data/policies/parent_call.cas
```diff
@@ -0,0 +1,15 @@
+virtual resource bar {
+	fn read(domain source) {
+		allow(source, this, file, read);
+	}
+}
+
+resource foo inherits bar {
+    fn read(domain source) {
```
dburgener, line 8: This line and line 14 have spaces, while the rest of the rest of the file has tabs.

data/error_policies/parent_call.cas
```diff
@@ -0,0 +1,15 @@
+virtual resource bar {
+	fn read(domain source) {
+		allow(source, this, file, read);
+	}
+}
+
+resource foo inherits bar {
+    fn read(domain source) {
```
dburgener, line 8: This line and line 14 have spaces, while the rest of the rest of the file has tabs.

dburgener:
Posted some initial comments.  As you mentioned, the syntax here is likely to change.  We'll try to get a final decision on that soon.

src/parser.lalrpop
```diff
@@ -127,7 +127,10 @@ Stmt: Statement = {
 }
 
 StmtBody: Statement = {
-	<c:(<TypeName> ".")?> <n:NameDecl> "(" <a:Comma<CastArg>> ")" => Statement::Call(Box::new(FuncCall::new_with_casts(c, n, a))),
+	<c:(<TypeName> ".")?> <n:NameDecl> "(" <a:Comma<CastArg>> ")" => Statement::Call(Box::new(FuncCall::new_with_casts(c, None, n, a))),
+	//<c:(<TypeName> "<")> <p:(<TypeName> ">.")> <n:NameDecl> "(" <a:Comma<CastArg>> ")" => Statement::Call(Box::new(FuncCall::new_with_casts(Some(c), Some(p), n, a))),
+	//<c:(<TypeName> ".")> <p:(<TypeName> "::")> <n:NameDecl> "(" <a:Comma<CastArg>> ")" => Statement::Call(Box::new(FuncCall::new_with_casts(Some(c), Some(p), n, a))),
+	<c:(<TypeName> ".")?> <p:NameDecl> "::" <n:NameDecl> "(" <a:Comma<CastArg>> ")" => Statement::Call(Box::new(FuncCall::new_with_casts(c, Some(p), n, a))),
```
matt-sheets, line 133: This is not going to be the final solution.  Do not merge until this is fixed.

data/expected_cil/parent_call.cil
```diff
@@ -0,0 +1,151 @@
...
+(typeattributeset resource (foo))
+(type xyz)
+(roletype object_r xyz)
+(typeattributeset abc (xyz))
+(typeattributeset resource (xyz))
+(macro abc-read ((type this) (type source)) (allow source this (file (read))))
+(macro bar-read ((type this) (type source)) (allow source this (file (read))))
+(macro foo-read ((type this) (type source)) (allow source this (dir (read))))
+(macro xyz-read ((type this) (type source)) (call abc-read (abc source)))
+(call foo-read (bar dom))
```
dburgener, line 143: Discussed on call.  I think this should probably be "(call bar-read (foo dom))" instead.

src/functions.rs
```diff
@@ -1993,12 +2055,29 @@ impl ValidatedCall {
...
+                    tmp_str.replace(&class_name.to_string(), cast_name.to_string().as_ref())
+                } else {
+                    tmp_str
+                }
+            }
+            None => {
+                // If we have a cast name but no class name that means we are really
+                // looking for the cast's function so append the cast's name
+                // to the function to find the correct one.
+                if let Some(cast_name) = &call.cast_name {
```
dburgener, line 2075: There's a top level get_cil_name() in ast.rs that does this logic.

src/functions.rs
```diff
@@ -2024,10 +2103,32 @@ impl ValidatedCall {
...
+                    Some(call),
+                    types,
+                    context,
+                    Some(file),
+                )?;
+            }
+        }
+
+        // All unwraps here are safe due to if checks
+        let mut args = if call.class_name.is_none() && call.cast_name.is_some() {
```
dburgener, line 2120: You should be able to write this more cleanly as a match statement I think.  Something like:

```
let mut args = match (call.class_name, call.cast_name) {
    (None, Some(cast_name)) => { ... },
    (None, None) => { ... },
    (Some(class_name), _) => { ... },
}
```

dburgener: Of course, rewriting it that way really suggests a further rewrite to:

```
let mut args = match call.class_name {
    Some(class_name) => { ... },
    None => { match call.cast_name {
        Some(cast_name) => { ... },
        None => { ... },
    }
}
```

which might make the intent cleaner.  I'm not sure I've actually followed the logic of this part yet though.

data/policies/parent_call.cas
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+resource xyz inherits abc {
+	fn read(domain source) {
+		abc.read(source);
+	}
+}
+
+// TODOs
```
dburgener, line 29: Are there TODOs in the code for what's left to do functionality wise to support these use cases?

dburgener (changes requested):

src/functions.rs
```diff
@@ -1770,6 +1770,68 @@ impl fmt::Display for FunctionArgument<'_> {
...
+// Validate that the parent provided both exists and is actually a parent of the current resource.
+fn validate_inheritance(
+    call: &FuncCall,
+    class_info: Option<&TypeInfo>,
+    parent_name: &CascadeString,
+    file: Option<&SimpleFile<String, String>>,
+) -> Result<(), ErrorItem> {
+    match class_info {
+        Some(class_info) => {
+            // In the case where we <parent>func it will look like we are our own parent and thats fine
```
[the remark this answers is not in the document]
matt-sheets: This was actually with the old parsing.  I am going to delete this comment for now since I don't think it is accurate anymore.

data/policies/parent_call.cas
matt-sheets: The TODO added in one of your previous comments should cover both of these.

matt-sheets: But this does bring up another question.  Do we want <bar>.read(this); to work with this PR?  Currently this is broken at the parser level.

src/functions.rs
```diff
@@ -1770,6 +1770,68 @@ impl fmt::Display for FunctionArgument<'_> {
...
+            .map(|ti| ti.is_setype(types))
+            .unwrap_or(false)
+    {
+        return Err(ErrorItem::make_compile_or_internal_error(
+            "Cannot typecast",
+            file,
+            s.get_range(),
+            "This is not something that can be typecast",
+        ));
+    } else if let Some(cast_type_info) = cast_ti {
```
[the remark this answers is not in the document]
matt-sheets: cast_ti is also None for the validate_argcast stuff, since it was not originally used.  So right now, being None is okay.

data/policies/parent_call.cas
dburgener: What do you mean by ".read(this)"?  Not sure I follow the question.

matt-sheets: Sorry markup ate the comment see edit above

dburgener: That makes more sense, but I'm still not sure I follow.  What behavior is `<bar>.read(this);` supposed to enable?  In the test case it looks like it should just be `xyz.read(this);` without brackets?  What's the difference between `<bar>.read(this);` and `bar.read(this)`?

src/functions.rs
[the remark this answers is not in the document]
dburgener: I think we should pass cast_ti through validate_argcast from validate_argument.  It looks like validate_argument checks the validity of the cast target directly.  That logic should probably move into validate_cast and apply to both situations?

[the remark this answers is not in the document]
dburgener: It looks like all we're using the TypeInstance for is to check if it matches TypeInstance::SEType.  That should be equivalent to calling is_setype() on the TypeInfo, I think?

Ideally, casting should work without any need to rely on TypeInstances, since we care what these things *are* not what they're *called*.

data/error_policies/parent_call.cas
```diff
@@ -0,0 +1,24 @@
+virtual resource bar {
+	fn read(domain source) {
+		allow(source, this, file, read);
+	}
+}
+
+virtual resource zap {
+	fn write(domain source) {
+		allow(source, this, file, read);
```
dburgener, line 9: Doesn't overly matter to the point of the test, but the write() function should probably grant the write permission?

src/ast.rs
```diff
@@ -421,19 +426,33 @@ pub struct FuncCall {
 
 impl FuncCall {
     pub fn new_with_casts(
-        cn: Option<CascadeString>,
+        cn: Option<(CascadeString, Option<CascadeString>)>,
         n: CascadeString,
         a: Vec<(Argument, Option<CascadeString>)>,
     ) -> FuncCall {
+        if let Some(some_cn) = cn {
```
dburgener, line 433: This "if let" followed by handling for the None case really should be a match block.

src/functions.rs
```diff
@@ -1772,6 +1772,96 @@ impl fmt::Display for FunctionArgument<'_> {
     }
 }
 
+// This seems to be a false positive on clippy's part
+// When I remove the return things don't compile correctly
```
dburgener, line 1776: Are you sure you're removing the correct return?  The one on 1828 does look needless to me.

src/functions.rs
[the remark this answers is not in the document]
dburgener: Looking at the latest version of the code, I think the question about cast_ti being None still stands.  The validate_argcast case has become guaranteed to be Some(), and the ValidatedCall case is the return of types.get().

Without even thinking about callers though, I'm not sure how doing any casting when the target is None makes sense, so it should probably be an error.

src/functions.rs
```diff
@@ -1772,6 +1772,96 @@ impl fmt::Display for FunctionArgument<'_> {
...
+) -> Result<(), ErrorItem> {
+    match class_info {
+        Some(class_info) => {
+            if !class_info.inherits.contains(parent_name) && class_info.name != parent_name.as_ref()
+            {
+                return Err(ErrorItem::make_compile_or_internal_error(
+                    "Invalid Parent",
+                    file,
+                    call.get_name_range(),
+                    "Resource does not inherit from given parent",
```
dburgener, line 1849: It might be a domain too, right?  Probably should say "type does not..."?

src/functions.rs
```diff
@@ -2009,12 +2099,20 @@ impl ValidatedCall {
...
-                match types.get(convert_class_name_if_this(class_name, parent_type)?.as_ref()) {
+                let tmp_str = match types
+                    .get(convert_class_name_if_this(class_name, parent_type)?.as_ref())
+                {
                     Some(type_name) => get_cil_name(Some(&type_name.name), &call.name),
                     None => call.get_cil_name(), // Expected to error out below
+                };
+                // If we have a cast name we really want to call that instead.
+                if let Some(cast_name) = &call.cast_name {
+                    tmp_str.replace(&class_name.to_string(), cast_name.to_string().as_ref())
```
dburgener, line 2110: Could this cause a problem if the function name contains the class name as a substring?  In other words, if I have a resource named foo, with a function named foobar().  Then the cil name is foo-foobar(), and if I cast to baz, this replace, would turn it into the invalid baz-bazbar() instead of the intended baz-foobar().

src/functions.rs
```diff
@@ -2040,10 +2138,29 @@ impl ValidatedCall {
...
+            }
+        }
+
         let mut args = match &call.class_name {
-            Some(c) => vec![convert_class_name_if_this(c, parent_type)?.get_cil_name()],
-            None => Vec::new(),
+            Some(class_name) => {
+                vec![convert_class_name_if_this(class_name, parent_type)?.get_cil_name()]
+            }
+            None => match &call.cast_name {
```
dburgener, line 2158: I feel like this section of code is a little inconsistent about what it assumes about the broader codebase.  I *think* that based on the current state of ast.rs it's impossible at the parser level for class_name to be None while cast_name is Some().  The block starting on 2141 seems to assume that, by not handling the class_name None case, which is reasonable enough.  But now we have special handling for is class_name is None and cast_name is Some(), which I justified the above block as being fine because that's impossible.

I think it is impossible for class_name to be None while cast_name is Some(), but I think this function should either 1. Assume that and act accordingly or 2. Verify that and return an internal error if it's false.   If 1, that should probably be documented in a comment.

src/functions.rs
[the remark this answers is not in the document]
matt-sheets: The None case for cast_ti should be handled at line 1824 in functions.rs.

dburgener (approved):

dburgener:
Merged, thanks!

PR #87, opened 2023-01-05
Add support for portcon() built in function

dburgener:
Also includes support for port numbers as a built in type and stubbing for ip address types

A notable omission at this point is that let bindings don't yet work with ports or ip addresses.

For example:

```
let my_port = 1234;

portcon("tcp", my_port, some_context);
```

doesn't work yet.

That will be addressed as future work


src/internal_rep.rs
```diff
@@ -1539,6 +1541,121 @@ fn call_to_fc_rules<'a>(
...
+        validate_arguments(c, &target_args, types, class_perms, context, Some(file))?;
+    let mut args_iter = validated_args.iter();
+
+    let proto = args_iter
+        .next()
+        .ok_or_else(|| ErrorItem::Internal(InternalError::new()))?
+        .get_name_or_string(context)?;
+    let proto = match proto.as_ref() {
+        "\"tcp\"" => Protocol::Tcp,
+        "\"udp\"" => Protocol::Udp,
```
matt-sheets, line 1618: It may be nice to lowercase the proto string to support stuff like TCP or Tcp

data/policies/networking_rules.cas
```diff
@@ -0,0 +1,7 @@
+resource my_port {
+	portcon("tcp", 1234);
```
matt-sheets, line 2: I would add an error case test make sure we are erroring correctly on tcp/udp.  Also do you think it's worthwhile to check that the port is in the valid port range?

src/internal_rep.rs
dburgener: My inclination is to say that we don't really want to support really perverse stuff like "tCP" or something.  Feels like that would make the policy less maintainable.  Maybe just enumerating "tcp" and "TCP" would be better than the case insensitive compare?

matt-sheets: I am good with just TCP or tcp.

PR #93, opened 2023-01-13
Fix bug where context lookups discarded the range

### LINKED ISSUE #92: get_name_or_string() returning CascadeString with Range of None
We believe that the code [here](https://github.com/dburgener/cascade/blob/main/src/internal_rep.rs#L2752 ) is erroneously stripping the range of the cascade string.

dburgener:
Previously context lookups were done on string slices instead of CascadeStrings, meaning that returning the looked up value required constructing a new CascadeString with None range.  Instead, have them take CascadeStrings, so if the lookup fails, we can just return the full CascadeString with its range.

closes #92

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -116,11 +116,11 @@ impl<'a> Context<'a> {
     // returns None
     // If the object is not in the context, then the string is valid elsewhere,
     // so just return the string
-    pub fn get_name_or_string(&self, arg: &str) -> Option<CascadeString> {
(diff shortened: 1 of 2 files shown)


matt-sheets (approved):

matt-sheets:
Do you think it's worthwhile for me to test this on the fs_context branch first before this is merged?

dburgener:
Yeah, that's a good idea.  Let me know how that testing goes.

matt-sheets:
This fixed the issue on fs_context for me.

PR #94, opened 2023-01-19
[RFC] Split up internal_rep.rs

dburgener:
Internal_rep.rs has grown to be a large catch-all file.  The code base would be more logically organized if it were split up into parts.

This PR creates three new modules pulled out of internal_rep.rs: functions, machine and alias_map.  Machine and alias_map are low hanging fruit - an essentially disconnected set of functionality that should obviously be a separate module, but historically grew within internal_rep.rs.  Functions is the heavier lift in terms of changes, but also the big win in terms of reducing the mess that is internal_rep.rs.  The new functions.rs contains all the generic function code, as well as the code for the built-in functions.

The mildly convoluted git history on this PR is very intentional to preserve git history on each of the files.  The basic method is to `git mv` internal_rep.rs to a temporary location, then on a separate branch, also `git mv` it to the new file.  These branches then have a merge conflict and on resolution we keep both the newly renamed file (eg machine.rs) and the temporarily renamed tmp.rs.  The result is that git recognizes each of these files as being a rename to internal_rep.rs when calculating history.  We then `git mv` tmp.rs back to internal_rep.rs to restore the repo with a newly created copy of internal_rep.rs alongside the original state, with both internal_rep.rs and the new file having the git history of internal_rep.rs due to the convoluted moves.

Finally, we delete the contents of our new file that do not belong in it, and the contents we're moving to the new file out of internal_rep.rs, and we adjust our internal includes (nothing about this commit affect public facing includes).

After all is said and done, we have separated files, and this history with moves and merges allows git to trace the history of each file.  `git blame` should work as expected.  `git log` requires the `--follow` argument in order to follow moves.

There is probably more splitting of internal_rep.rs that could be done.  Ideally, I'd like to eliminate the file entirely.  But this seems like a good starting place to me.  Functions.rs is the challenge to rename, since it is so high touch, so I grabbed a moment when there are few outstanding PRs on the repo to do this.  Remaining moves should be easier since with this much splitting, PRs should generally touch internal_rep.rs a lot less.


matt-sheets (approved):
Things look reasonable.  I have pulled the branch and did some minor testing of it.  The git history looks maintained in the new files.

The only comment I have is we can probably break down functions.rs further.  Portcon fscontext filecontext etc, could live in their own files.  But I am 100% okay with leaving well enough alone for now.

dburgener:
Yeah, I hear you on breaking up functions.rs.  Once I pulled that out, I realized that I'd essentially moved the problem, since now functions.rs is 2.6k lines.

I do think that that needs to be future work though in order to not be too disruptive to our primary functional goals while doing this sort of cleanup.

I do think that there's a lot of room to improve how we handle built-ins though.  There's plenty of code reuse that could be handled cleaner with helpers.  If we add the right helpers, the built-ins become a lot smaller, and somewhat stop justifying their own files.  Possibly a split like "functions.rs" for the generic portions and "builtins.rs" for the built in functions would be a sensible split.  But like I said, future work I think.

l0kod:
I like this kind of split, but I'm not a fan of such merges because they make the repository non-bisectable, which is from my point of view one of the most important property to have. I'll take a closer look tomorrow.

l0kod:
We can do both by changing the internal_rep inclusion/module on other files when renaming internal_rep.rs in a way that makes everything work the same way, but I'm not sure it is worth it compared to creating partial copies of files (for which git blame will identify this commit as the first one).

dburgener:
I don't follow that second suggestion. Can you elaborate on what you mean?

dburgener:
I just tried a git bisect across this series of changes and it worked fine.  Are you sure that this will make the repo non-bisectable?

dburgener:
Based on this discussion, we're doing this the other way https://github.com/dburgener/cascade/pull/98

PR #101, opened 2023-01-27
Run PR failing style checks only on a known good toolchain version

dburgener:
Clippy regularly updates with new checks.  When this happens, we update our repo to account for this (either fixing the code, or disabling the checks).  However, this often hits a new PR that happens to be the first PR to run against the new check.  That PR doesn't necessarily introduce an issue, but fails anyways.

Split the style checks to run against an older known good version of the toolchain, which can fail PRs, and also informationally run against the latest toolchain.  This way, PRs won't block on new clippy checks, but we'll still be notified of the need to fix them.

Once we fix the latest clippy checks, we can update the known good version to the new clippy version and continue checking against that version.

Clippy is the specific example here, but this updates the toolchain for the full checks step (except building nightly) so that we're running against a stable known good for all style checking.

### DIFF
--- a/.github/workflows/clippy.sh
+++ b/.github/workflows/clippy.sh
@@ -2,6 +2,8 @@
 
 set -e -u -o pipefail
 
-rustup run stable cargo clippy -- -A clippy::new_without_default --deny warnings
+VERSION=${1:-stable}
 
-rustup run stable cargo clippy --tests -- -A clippy::new_without_default -A clippy::expect_fun_call
+rustup run ${VERSION} cargo clippy -- -A clippy::new_without_default --deny warnings
+
+rustup run ${VERSION} cargo clippy --tests -- -A clippy::new_without_default -A clippy::expect_fun_call
--- a/.github/workflows/doc.sh
+++ b/.github/workflows/doc.sh
@@ -2,4 +2,6 @@
 
 set -e -u -o pipefail
 
-rustup run stable cargo doc --no-deps
+VERSION=${1:-stable}
+
+rustup run ${VERSION} cargo doc --no-deps
--- a/.github/workflows/format.sh
+++ b/.github/workflows/format.sh
@@ -2,4 +2,6 @@
 
 set -e -u -o pipefail
 
-rustup run stable cargo fmt --all -- --check
+VERSION=${1:-stable}
+
+rustup run ${VERSION} cargo fmt --all -- --check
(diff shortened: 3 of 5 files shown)


dburgener:
I think I've done this right, but github just doesn't support this use case: https://github.com/orgs/community/discussions/15452

In short, the desired outcome here is to run an informational check (clippy stable) that doesn't fail CI, but does inform us.  This approach sort of works, the github actions run (https://github.com/dburgener/cascade/actions/runs/4026294806) succeeded.  Our branch settings don't require checks to pass, so it doesn't block the PR.  But this commit is still getting a red x rather than a green checkmark.  As mentioned in the post above, that seems to be the expected behavior of this approach currently, and there isn't currently a straightforward alternative to get the behavior we want.  Possibly something like setting the test to always pass, but post a comment on the PR if it would fail is a reasonable approach.

dburgener:
According to this comment: https://github.com/actions/runner/issues/2347#issuecomment-1367587399 step level continue-on-error at least makes the PR UI report as passing.  I don't think it's clear in that scenario that anything failed though, so it's not an ideal solution.  I'm going to try it out and see how it looks.

dburgener:
Yeah, as expected that version gives full green checkmarks, but it's impossible to see that anything has failed.  Possibly combining that with something like this: https://github.com/mainmatter/continue-on-error-comment would be a satisfactory solution.

dburgener:
Okay, I think this is ready.  I decided that the trust model is cleaner using only official github actions, and it's not that hard to add a comment action using that.  So I've detected failure of the stable clippy and added a comment there directing the maintainer to look at the logs.  It's not perfect, but it gets the job done.

The tests are failing, but at first glance that looks like an infrastructure issue.  This PR shouldn't have affected them at all, and they're passing on my box after a rustup update, both on this branch and main.  I'll dig into them more in a bit.

dburgener:
Looks like the test failure is related to miniz_oxide v0.6.3.  This is an indirect dependency of cascade via flate2.  They published version 0.6.3 about an hour ago: https://crates.io/crates/miniz_oxide/versions, so it isn't getting picked up on my box yet (even with a rustup update I'm on 0.6.2), but github has 0.6.3.  It seems as though there is an assertion failure in the new version.

dburgener:
https://github.com/Frommi/miniz_oxide/commit/dd2fa3e33c7d8ce1bb8bbd7329a0dd571f4e2df6#r98948840

dburgener:
Looks like a fix: https://github.com/Frommi/miniz_oxide/pull/134

matt-sheets (approved):

dburgener:
According to crates.io, miniz_oxide 0.6.4 is out, which should address this issue.  I'm rerunning tests.

PR #102, opened 2023-01-30
Address clippy 1.66 linting suggestions

dburgener:


dburgener:
I'd like to resolve and merge https://github.com/dburgener/cascade/pull/101 before taking this, since having this issue helps for testing #101.  Once #101 merges, these sorts of issues become less urgent and can be addressed as convenient.

l0kod (approved):

PR #104, opened 2023-02-01
Run all the CI against nightly.  Don't hard fail nightly test runs

dburgener:


dburgener:
Realized some issues with this.  I've converted it back to draft to work on it more.

dburgener:
Okay, this is ready now.  We don't need to test all the combinations of secilc versions and stable vs nightly, since we're already checking that we generate identical CIL for all of our positive test cases.  Just running one round of tests on nightly to ensure that everything is good there is sufficient.

matt-sheets (approved):
Looks good.  Where you able to get the workflow to fail in a way to generate the comment for a smoke test?

dburgener:
>  Where you able to get the workflow to fail in a way to generate the comment for a smoke test?

No, I just copied the code from the clippy failures, which I was able to test.  That's a good point.  I'll instrument it to fail and try a push.

PR #105, opened 2023-02-01
Allow context nesting

dburgener:
The symbol table should nest within blocks and do shadowed lookups.  This enables the removal of the special handling for global bindings, which was pretty broken.

### DIFF
--- a/data/policies/let.cas
+++ b/data/policies/let.cas
@@ -8,8 +8,7 @@ let rw_file_perms = [ read_file_perms write ];
 
 resource bar {}
 
-// TODO: re-add once global binding is using Contexts
-//let baz = bar;
+let baz = bar;
 
 domain foo {
 
@@ -21,5 +20,5 @@ domain foo {
 	allow(foo, bar, file, binding_with_annotation);
 	allow(this, bar, file, internal_binding);
 	allow(this, bar, file, nested_binding);
-	//allow(foo, baz, file, write);
+	allow(foo, baz, file, write);
 }
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -249,7 +249,6 @@ pub fn get_global_bindings(
     classlist: &mut ClassList,
     file: &SimpleFile<String, String>,
 ) -> Result<(), CascadeErrors> {
-    let tm_clone = &types.clone(); // Temporary workaround to get it compiling
     for e in &p.policy.exprs {
         if let Expression::Stmt(Statement::LetBinding(l)) = e {
             let let_rvalue = ArgForValidation::from(&l.value);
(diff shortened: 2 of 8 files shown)


dburgener:
Highly recommend reviewing by commit.  It's a lot of changes, but should be well split up by commit.

data/policies/let.cas
```diff
@@ -8,8 +8,7 @@ let rw_file_perms = [ read_file_perms write ];
 
 resource bar {}
 
-// TODO: re-add once global binding is using Contexts
-//let baz = bar;
+let baz = bar;
```
matt-sheets, line 11: I am guessing bar has to actually exist here.  Maybe worthwhile adding an error policy to check that.

src/compile.rs
```diff
@@ -1574,8 +1546,10 @@ pub fn call_derived_associated_calls<'a>(
                         }) {
                             let call = make_associated_call(resource_name, f);
                             let args = vec![FunctionArgument::new_this_argument(t)];
+                            // TODO: Should there be a parent_context here?  I think this is
+                            // effectively a "fake" context since we're not really parsing the tree
                             let mut local_context =
-                                BlockContext::new(BlockType::Domain, types, Some(t));
+                                BlockContext::new(BlockType::Domain, Some(t), None);
```
matt-sheets, line 1552: I am going to ask a question that will hopefully help get to the bottom of this todo.  Can we have chains of associated calls and inheritance?  Something like iptables_tmp inherits tmp_file, tmp_file has some associated calls, tmp_file inherits from any_file, any_file also has some associated calls.  If iptables associates with iptables_tmp, would the inheritance between tmp_file and any_file come into play here?  If it would we may need to include the parent of the thing we are associating with.

src/compile.rs
```diff
@@ -1630,18 +1605,38 @@ fn do_rules_pass<'a>(
...
+                        types,
+                        file,
+                    ) {
+                        Ok(()) => (),
+                        Err(e) => errors.append(e),
+                    }
+                }
+                // TODO: The note in ValidatedStatement::new() says "TODO: This is where local
+                // scope let bindings should happen".  Isn't that what we just did above?  Is there
+                // still a TODO here?
```
matt-sheets, line 1631: I don't think so.  The logic that was in ValidatedStatement was moved here, and the other logic added in this PR seems to be handling the let statements correctly.

src/compile.rs
dburgener: > Can we have chains of associated calls and inheritance?

Yes.

>  If it would we may need to include the parent of the thing we are associating with.

Do we?  I probably need to think about this scenario more.  Just looking at the immediate context, it looks like all we're setting up is that "this" is a symbol for whatever type is associating here. I guess the real question is whether we'll need to access any bindings (and what scope is even relevant for those bindings?).  My gut says that because we're generating all the call args internally that there can't be any bindings in play here?

I'm not 100% sure though.  Happy to sit down and talk through this case sometime.

matt-sheets: I guess the proof is in the pudding, if things are working how we want them to this is probably fine as is.  But I am good with bouncing it around if you think we need to.

matt-sheets (approved):

PR #106, opened 2023-02-03
[RFC] Add parser support for drop keyword

dburgener:


dburgener:
Just putting this up for tracking and comment.  Two major issues need resolution before its ready to merge:

1. Right now this doesn't throw a warning that drop isn't supported yet, but it should
2. We haven't actually picked a keyword yet.  I like "drop" and used it here, but that's very changeable.

dburgener:
> We haven't actually picked a keyword yet. I like "drop" and used it here, but that's very changeable.

We discussed this yesterday, and everyone agreed that they liked "drop".  There is also agreement that we won't actually support "drop" until the CIL side deny work (https://lore.kernel.org/selinux/20221215213429.998948-1-jwcart2@gmail.com/T/#t) merges and is in a release.  At that point we'll definitely need to add "target libsepol version" support, and only support drop on new selinux userspace versions.

The other two notes (throw a warning, and documentation) still need addressing, so I'll keep this in RFC state for now.  I'm working on a PR to add warning support generally which I hope to publish this week.  Once that lands, this can rebase on top of it and add the warning for drop.

dburgener:
All above-noted todos are now addressed.  Marking this as ready for review.

doc/TE.md
```diff
@@ -208,32 +208,85 @@ For more details and examples, see resource_association.md
...
+virtual domain bar inherits foo {
+	drop allow(this, self, capability, sys_admin);
+}
+
+domain baz inherits bar {}
+```
+
+In the above example, `foo` gets both the `sys_admin` and `net_admin`
+capabilities; `bar` inherits the `net_admin` capability, but drops the
+`sys_admin` capability so doesn't have that permission.  `baz`, inheriting
```
matt-sheets, line 268: "so it doesn't" read a bit better.

doc/TE.md
```diff
@@ -208,32 +208,85 @@ For more details and examples, see resource_association.md
...
+virtual resource bar {}
+resource baz inherits bar {}
+
+domain foo {
+	allow(this, bar, file, [ read write append ]);
+	drop allow(this, baz, file, append);
+}
+```
+
+In the above example, foo can generally read, write and append files with `bar`
```
matt-sheets, line 284: Drop "generally"

matt-sheets (approved):

PR #110, opened 2023-02-09
Add comment workflow

matt-sheets:
With previous updates to the clippy workflow
there was attempts to add a comment to the PR
if clippy failed.  This was done under pull_request
action.  This would only work on PRs from the owner
of the repo.  With this comment the comments will
be left under the workflow_run action.

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -16,7 +16,7 @@ jobs:
         # regularly update the known good once we know that tests
         # pass on it
         # This needs to stay in sync with rust.yml
-        rust-toolchain: [ 1.66, stable, nightly ]
+        rust-toolchain: [ 1.67, stable, nightly ]
     if: >
       github.event.workflow_run.event == 'pull_request' &&
       github.event.workflow_run.conclusion == 'success'
@@ -44,7 +44,7 @@ jobs:
       - run: unzip clippy-${{ matrix.rust-toolchain }}.zip
 
       - name: Notify clippy failure
-        if: (matrix.rust-toolchain != '1.66')
+        if: (matrix.rust-toolchain != '1.67')
         uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -9,6 +9,14 @@ on:
 jobs:
   upload:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # We lock our failing tests on a known good version and
+        # informationally check the latest as well.  We should
+        # regularly update the known good once we know that tests
+        # pass on it
+        # This needs to stay in sync with rust.yml
+        rust-toolchain: [ 1.66, stable, nightly ]
     if: >
       github.event.workflow_run.event == 'pull_request' &&
       github.event.workflow_run.conclusion == 'success'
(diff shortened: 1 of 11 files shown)


.github/workflows/rust.yml
```diff
@@ -38,17 +39,19 @@ jobs:
       id: clippy
       run: .github/workflows/clippy.sh ${{ matrix.rust-toolchain }}
       continue-on-error: ${{ matrix.rust-toolchain != '1.66' }}
-    - name: Notify clippy failure
-      if: ${{ (steps.clippy.outcome != steps.clippy.conclusion) && (matrix.rust-toolchain != '1.66') }}
-      uses: actions/github-script@v6
+    - name: Store clippy flag
```
dburgener, line 42: I believe the nightly tests need the same behavior as this?

.github/workflows/comment.yml
```diff
@@ -0,0 +1,62 @@
...
+  upload:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # We lock our failing tests on a known good version and
+        # informationally check the latest as well.  We should
+        # regularly update the known good once we know that tests
+        # pass on it
+        # This needs to stay in sync with rust.yml
+        rust-toolchain: [ 1.66, stable, nightly ]
```
dburgener, line 19: Can't 1.66 not generate the flag, so we don't need it here?

.github/workflows/comment.yml
```diff
@@ -0,0 +1,62 @@
+name: Comment on clippy
+
+on:
+  workflow_run:
+    workflows: ["Rust"]
+    types:
+      - completed
+
+jobs:
+  upload:
```
dburgener, line 10: Why is this named "upload"?  Isn't it doing a download?

.github/workflows/comment.yml
```diff
@@ -0,0 +1,62 @@
...
+        # regularly update the known good once we know that tests
+        # pass on it
+        # This needs to stay in sync with rust.yml
+        rust-toolchain: [ 1.66, stable, nightly ]
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: 'Download artifact'
+        uses: actions/github-script@v3.1.0
```
dburgener, line 25: I doubt we want to pin on either that specific or old of a version?  If you copied this from the docs/example, it's probably just old and out of date.  Can you check what the latest github-script version is and use that?  (Same with upload-artifact in the other file)

.github/workflows/comment.yml
dburgener: By which I mean, I guess 1.66 will upload an artifact.  but we don't comment there, so why not just not run this against 1.66 and not use the artifact?

dburgener (changes requested):

.github/workflows/comment.yml
matt-sheets: From the example, but I can name it whatever.

.github/workflows/rust.yml
matt-sheets: This will run with the nightly tests in the matric unless you are referring to something else.

dburgener: I mean the "notify nightly test failure" step on line 89.

.github/workflows/comment.yml
dburgener: The comment is a little weird now, since 1.67 will move in rust.yml, but is explicitly mentioned in the comment here.  Maybe s/1.67/"the known good clippy version"/?

dburgener:
Merged, thanks!

PR #115, opened 2023-02-13
Fix clippy warning

matt-sheets:


matt-sheets:
Apparently, this broke a lot more than I thought it would.

dburgener:
Yeah, I think I've had an initial failure on infra issues on every PR run I've done today.  Seems like there's some sort of health issue going on.

src/lib.rs
```diff
@@ -405,6 +405,9 @@ mod tests {
         assert!(count > 9);
     }
 
+    // assert! has a format! macro in it, clippy gets confused about this.
```
dburgener, line 408: If assert!() doesn't follow uninlined_format_args, we should probably just add a global exception in .github/workflows/clippy.sh instead of these specific exceptions.

dburgener (approved):

dburgener:
Looks like it has some conflicts.  Can you fix before I merge?

matt-sheets:
> Looks like it has some conflicts. Can you fix before I merge?

Should be good now.

dburgener:
Merged, thanks.

Issue #76, opened 2022-11-30
Error messages for object class permissions with a "2" are reported against the "2" version of the class


dburgener:
If you do something like:

```
allow(this, this: resource, capability, foo);
```

Then you will get a message that "foo" is not in the "capability2" object class.  The expectation is that the error message should refer to the "capability" object class instead.

PR #120, opened 2023-02-14
Bad Perms report for correct object class

matt-sheets:
Previously on object classes with "2" (e.g. capability2) if a permission was not part of that object class the compiler would always report the 2nd object class.  Now the correct object class from the cascade policy should be reported.

Issue #76

### DIFF
--- a/src/internal_rep.rs
+++ b/src/internal_rep.rs
@@ -866,6 +866,7 @@ impl<'a> ClassList<'a> {
         permission: &CascadeString,
         context: &BlockContext<'_>,
         file: &SimpleFile<String, String>,
+        original_class: Option<&CascadeString>,
     ) -> Result<(), ErrorItem> {
         let resolved_class = context.get_name_or_string(class);
         let class = resolved_class.as_ref().unwrap_or(class);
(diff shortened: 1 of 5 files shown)


src/internal_rep.rs
```diff
@@ -866,6 +866,7 @@ impl<'a> ClassList<'a> {
         permission: &CascadeString,
         context: &BlockContext<'_>,
         file: &SimpleFile<String, String>,
+        original_class: Option<&CascadeString>,
```
dburgener, line 869: Not overly wild about adding this aspect, as it seems like it exposes an internal implementation detail to callers unnecessarily.

Without having thought about it too much, would it work to check for the "2" versions in the error message and just convert back based on the internal mapping between originals and "2"s?  Or do we need to know that we're in the recursive lookup at that point?

matt-sheets: It depends on how we want to present the error to the user.  With this solution we will tell them non-2 and 2 correctly, which we need to know if we have entered the function recursively to check the 2s.

If we just always want to tell them the non 2s I can just simply map it back.

matt-sheets: From a cil, AVC, non-cascade standpoint I think it's better to tell them if it is a non-2 or a 2, but since we just combine them in cascade it may be okay to just map it back.

dburgener: I think my preference here is to create a new verify_permission_helper() to pass the recursive parameter to, rather than exposing it to callers.

dburgener (approved):

dburgener:
closes #76

dburgener:
Merged, thanks

Issue #60, opened 2022-10-20
`:` isn't lexed


dburgener:
`file_context("/dev/log", socket, system_u:object_r:devlog_t:mls_systemhigh);`

Reports "invalid" character on the ":".  At a quick glance, it looks like ":" is handled in context parsing, but not in the lexing. So perhaps just including this in the lexer will pass it on to the rest which is unit tested and *should* work.  On the other hand, maybe we want to parse contexts as contexts?  Needs a more comprehensive investigation.

dburgener:
This is more than just the lexer issue.  `call_to_fc_rules()` expects a *resource* here, which works for current unit tests expecting a resource, but isn't flexible to people providing full contexts.

This combined with `allow()` calls with targets that are `domain` makes me suspect that built-ins need a way to support optionally different argument types.

PR #121, opened 2023-02-15
Handle full contexts correctly

dburgener:
Previously, only bare types were accepted as contexts.  This change allows the specification of the full context.

### DIFF
--- a/data/policies/filecon.cas
+++ b/data/policies/filecon.cas
@@ -6,6 +6,7 @@ resource foo {
 	file_context("/etc/somesymlink", [lnk_file], this);
 	file_context("/var/somepipe", [fifo_file], this);
 	file_context("/var/somesocket", [sock_file], this);
+	file_context("/bin/some_bin", [file], system_u:object_r:foo:s0);
 	// Policies must include at least one av rule
 	allow(domain, foo, file, [read]);
 }
--- a/data/policies/filecon.cas
+++ b/data/policies/filecon.cas
@@ -7,6 +7,8 @@ resource foo {
 	file_context("/var/somepipe", [fifo_file], this);
 	file_context("/var/somesocket", [sock_file], this);
 	file_context("/bin/some_bin", [file], system_u:object_r:foo:s0);
+	file_context("/bin/some_bin2", [file], system_u:object_r:foo:s0-s0);
+	file_context("/bin/some_bin3", [file], system_u:object_r:foo:s0-s0:c0.c255);
 	// Policies must include at least one av rule
 	allow(domain, foo, file, [read]);
 }
--- a/src/parser.lalrpop
+++ b/src/parser.lalrpop
@@ -232,6 +232,14 @@ Arg: Argument = {
 	Quoted_String => Argument::Quote(<>),
 	Port => Argument::Port(<>),
 	IPAddr => Argument::IpAddr(<>),
+	Context => Argument::Var(<>),
+}
+
+Context: CascadeString = {
+	// TODO: don't discard the mls range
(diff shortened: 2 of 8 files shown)


dburgener:
closes #60

data/policies/filecon.cas
```diff
@@ -6,6 +6,7 @@ resource foo {
 	file_context("/etc/somesymlink", [lnk_file], this);
 	file_context("/var/somepipe", [fifo_file], this);
 	file_context("/var/somesocket", [sock_file], this);
+	file_context("/bin/some_bin", [file], system_u:object_r:foo:s0);
```
matt-sheets, line 9: From parser.lalrpop we are not using the mls right now, but it may be a good idea to add a file context with an mls range here as well to make sure we can parse that when its ready.

src/parser.lalrpop
```diff
@@ -227,6 +227,14 @@ Arg: Argument = {
 	Quoted_String => Argument::Quote(<>),
 	Port => Argument::Port(<>),
 	IPAddr => Argument::IpAddr(<>),
+	Context => Argument::Var(<>),
+}
+
+Context: CascadeString = {
+	// TODO: don't discard the mls range
+	<start: @L> <u: Symbol> ":" <r: Symbol> ":" <t: Symbol> <m: (":" <Symbol>)?> <end: @R> => {
```
matt-sheets, line 235: Just for my learning what are @L and @R?

dburgener: Well, I was hoping to link you to the official documentation, but instead I found this bug that they aren't documented :) https://github.com/lalrpop/lalrpop/issues/343  As far as I can tell the lalrpop project sadly seems abandoned, so hope of doc updates is low.

Anyways, what `@L` and `@R` is provide the location in the file of the left and right of the token next to them.  `@L` provides the leftmost position in the token to the right of it, and `@R` provides the rightmost position in the token to its left.  When used in conjunction, they're how we get our Ranges for CascadeStrings.

data/policies/filecon.cas
dburgener: Okay, very glad we decided to dig into this, because there's a problem with mls ranges.  The category portion of a label can be a comma-separated list.  So if we have something like:

```
some_call(system_u:object_r:type_t:s0-s0:c0,c1);
```

It's not clear if that's a function with one argument (`system_u:object_r:type_t:s0-s0:c0,c1`), or two arguments (`system_u:object_r:type_t:s0-s0:c0` and `c1`).

My first instinct is to say that we need to quote the full contexts or something, but that's a little annoying given that they contain elements declared elsewhere in the policy (eg types).

dburgener: Matt and I discussed this offline, and both think that allowing contexts to be optionally quoted, and requiring the quotes if they will contain commas, similar to how bash strings work with spaces is the most sensible approach here.

I've updated the PR to assume contexts can't contain commas, and the quote portion, along with helpful errors to point out that a comma in a context might need to be quoted will be future work.

data/policies/filecon.cas
```diff
@@ -6,6 +6,9 @@ resource foo {
...
+	file_context("/bin/some_bin", [file], system_u:object_r:foo:s0-s0);
+	file_context("/bin/some_bin", [file], system_u:object_r:foo:s0-s0:c0.c255);
```
matt-sheets, line 11: I am guessing at some point we want to at least warn if the paths are identical with different contexts for file cons.  I would update the paths to be different here just to future proof this.

dburgener: Good catch.  It looks like right now we silently drop the duplicates.  I think we should error there.  I filed #131 to investigate that.

matt-sheets (approved):

PR #123, opened 2023-02-16
Parse the "!" character in boolean expressions

dburgener:
Also expand boolean parsing test cases

### DIFF
--- a/data/policies/conditional.cas
+++ b/data/policies/conditional.cas
@@ -1,6 +1,7 @@
 resource foo {}
 
 let my_tunable = false;
+let my_tunable_2 = true;
 
 domain bar {
 	if (true) {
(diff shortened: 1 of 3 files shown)


data/policies/conditional.cas
```diff
@@ -17,6 +18,18 @@ domain bar {
 		allow(this, foo, file, getattr);
 	}
 
+	if (my_tunable && my_tunable_2) {
+		allow(this, foo, file, setattr);
+	}
+
+	if (my_tunable || my_tunable_2) {
+		allow(this, foo, file, open);
+	}
```
matt-sheets, line 27: If the parser can handle it, it would be a good idea to test stuff like (!(x && y)) and (x && !y)

matt-sheets (approved):

PR #127, opened 2023-02-21
Add parser support for optional blocks

dburgener:
We haven't defined exactly how optional policy will work yet, but all the details on the table involve optional blocks and refpolicy 3 makes use of optional blocks to retain information from refpolicy 2.

For now, just include all the optional rules in the policy with no special handling.  This will need to be fixed once we settle on the details of how optional works.


dburgener:
This might be a little premature, since we haven't worked out the full optional scheme yet.  But it nice to unblock refpolicy3 compile to at least have parser support on these, and I *think* we'll want it regardless, although the details of how their handled are up in the error.

src/ast.rs
```diff
@@ -417,6 +430,13 @@ impl Statement {
                     "The drop keyword can only be applied to function calls.".to_string(),
                 ))
             }
+            Statement::OptionalBlock(_) => {
+                return Err(ParseErrorMsg::new(
```
matt-sheets, line 434: I would add an error test for this.

src/parser.lalrpop
```diff
@@ -216,6 +217,10 @@ ElseBlock: Vec<Statement> = {
 	"else" "{" <Stmt+> "}" => <>
 }
 
+OptionalBlock: OptionalBlock = {
+	"optional" "{" <Stmt*> "}" => OptionalBlock::new(<>)
```
matt-sheets, line 221: I am guessing * is 0 to many.  Do we want to allow for an empty optional block?

dburgener: Two initial thoughts:

1. I think it might be desirable if someone wants to comment out all the rules in an optional block?
2. If we do want to ban them, the parser level may not be the best place.  It might make more sense to parse, and then report an error when actually processing the block in ValidatedStatement::new()

matt-sheets: 1. It would only be a few more lines to comment out the block itself.  So, I don't weigh that too heavily.
2. My initial thought would be less stuff to deal with past the parser, but I do agree the parser error is going to be less meaningful than one we put there ourselves.

One last thought is just that an empty optional block is meaningless.  Other empty blocks we have may not be.

I don't feel too passionately one way or another, so I will leave the call up to you.

src/ast.rs
dburgener: So, interestingly, it appears that this error is actually unreachable.  The parser permits drop before a let binding or function call, but at a parser level, drop cannot appear before an if block or an optional block.

It would be fairly straightforward to turn these parse errors into compile errors:

```
<start: @L> "drop" <end: @R> <mut s: StmtBody> ";"
```

Could just become:

```
<start: @L> "drop" <end: @R> <mut s:Stmt>
```

In addition to allowing drop in front of optional and if blocks, this would also allow the perverse "drop drop drop drop some_statement();" at a parser level, which seems mildly undesirable to me (I think it would be functionally identical to just one drop).

The advantage of allowing it at the parser, is that these error messages are more helpful than the "Unexpected word optional" that the parser returns here.  So it's probably worth letting this all pass through the parser and catching it here.

Thoughts?

matt-sheets: My first reaction is if we let it past the parser there may be weird corner cases we need to deal with later.  Like I didn't even think about stuff like "drop drop drop".  If you think it won't get out of hand, I am good with it going past the parser.

src/parser.lalrpop
dburgener: Having sat on this for a little while, what I would like to do is allow empty optional blocks.  I don't see it hurting anything, and I think the commenting out scenario is not unreasonable.  Perhaps it's something we might warn or lint on in the future.

As an aside, I'd generally prefer a more permissive parser with errors later as a non-strict general principle.  Since if we report errors after parsing, we can be more specific and helpful in the error message.

src/ast.rs
dburgener: Okay, I looked at this some more, and decided that really it makes more sense for drop to apply only to calls at a parser level, which simplifies the code in the ast for handling all of it, removing these error conditions.  I pushed that change as a new commit here, and I think it cleans all of this up and addresses this concern.

matt-sheets (approved):

PR #128, opened 2023-02-22
Allow non-parent functions to be cast

matt-sheets:
Previously one could only cast a function if it was from an inherited type.  Now you can cast a function as long as the function does not reference any associated types.


src/compile.rs
```diff
@@ -421,6 +421,75 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
     errors.into_result(())
 }
 
+// Go through all functions and check if they are castable
+// based only on their args
+pub fn initialize_castable(functions: &mut FunctionMap, types: &TypeMap) {
```
dburgener, line 426: This looks like it's checking the declared arguments in the function signature.  Is the relevant thing to check the arguments passed in a call?

data/policies/parent_call.cas
```diff
@@ -33,9 +33,30 @@ resource xyz inherits abc {
 	}
 }
 
-// TODOs once we add the ability to cast to
-// non-inherited things
-//
-//domain asd {
-//	asd<abc>.read(this);
-//}
+domain asd {
```
dburgener, line 36: Can you also test this situation?

```
resource foo {
   fn some_func(domain source, resource some_resource) {
       allow(source, some_resource, file, read);
   }
}

@associate([tmp])
domain bar {
    foo.some_func(this, this.tmp);
}
```

matt-sheets: That will compile in cascade but not with secilc.
```
Failed to resolve qwe.tmp in call argument list at out.cil:201
Failed to resolve call statement at out.cil:201
Failed to resolve AST
Failed to compile cildb: -2
```

dburgener: Interesting.  CIL shouldn't be seeing "qwe.tmp" at all.  It should have been converted to "qwe-tmp".

Sounds like we may have a bug where we aren't converting associated resource names in a call argument context?  Can you see if it reproduces with a simple call situation without casting involved and make a second PR to fix it?

Actually, looking again, this is a non-casting situation.  So your casting code isn't even running here.  I'm not actually sure if triggering the case I was thinking of is possible in the current code.  Let me give it some more thought...

data/error_policies/parent_call.cas
```diff
@@ -18,7 +18,29 @@ resource foo inherits bar {
...
+
+@associate([tmp])
+domain boo {
+	// Creates new resources boo.tmp and implicitly calls
+	// boo.tmp.associated_call_from_tmp(boo)
+	//
+	// boo.tmp inherits tmp
+
+	fn read_boo_tmp(domain source) {
+		allow(source, boo.tmp, file, [read]);
```
dburgener, line 44: Okay, I've figure out what I was trying to say in my other comment.  I want an error test where this call is not a built-in.  I'm concerned that that case may not error correctly in the current code.

matt-sheets: Something like this?
```
domain asd {
	asd<boo>.read_boo_tmp(boo); // Function is not castable
}

virtual resource tmp {
	@associated_call
	fn associated_call_from_tmp(domain source) {
		allow(source, tmp, file, [read]);
	}
}

@associate([tmp])
domain boo {
	// Creates new resources boo.tmp and implicitly calls
	// boo.tmp.associated_call_from_tmp(boo)
	//
	// boo.tmp inherits tmp

	fn read_boo_tmp(domain source) {
		boo.tmp.associated_call_from_tmp(source);
	}
}
```

dburgener: No, I mean passing an associated resource as an argument to a non-builtin.

```
domain asd {
        asd<boo>.some_function(some_resource);
}

virtual resource tmp {
	@associated_call
	fn associated_call_from_tmp(domain source) {
		allow(source, tmp, file, [read]);
	}
}

@associate([tmp])
domain boo {
	// Creates new resources boo.tmp and implicitly calls
	// boo.tmp.associated_call_from_tmp(boo)
	//
	// boo.tmp inherits tmp

	fn read_boo_tmp(domain source) {
		this.some_function(this.tmp);
	}
        fn some_function(resource res) {
                allow(this, res, file, read);
        }
}

File: data/policies/parent_call.cas (hunk 2 above)
Comment (dburgener) [inline data/policies/parent_call.cas:36]: The name of this file implies that it's only the "parent" version of casting.  I don't really mind having this here, but we should rename the file to call_casting or something like that.  I'd also be fine if you prefer putting this stuff in a new test file.

File: data/policies/parent_call.cas [hunk 4]
```diff
@@ -33,9 +33,30 @@ resource xyz inherits abc {
...
+	asd<abc>.read(this);
+	foo<abc>.read(this);
+}
+
+domain jkl inherits boo {
+	jkl<boo>.read_boo_tmp(this);
```
dburgener, line 42: Isn't the main thing we want to test the case when jkl is *not* a child of boo?

data/policies/parent_call.cas
dburgener: bump on this:

> Sounds like we may have a bug where we aren't converting associated resource names in a call argument context? Can you see if it reproduces with a simple call situation without casting involved and make a second PR to fix it?

matt-sheets: https://github.com/dburgener/cascade/issues/133

src/compile.rs
matt-sheets: We are marking some functions as non castable with this method so I am not sure if it is 100% useless.

data/policies/parent_call.cas
dburgener: Oh, yes.  I follow now.  Thanks for clarifying.

data/policies/call_casting.cas
```diff
@@ -33,9 +33,30 @@ resource xyz inherits abc {
...
+
+@associate([tmp])
+virtual domain boo {
+	// Creates new resources boo.tmp and implicitly calls
+	// boo.tmp.associated_call_from_tmp(boo)
+	//
+	// boo.tmp inherits tmp
+
+	fn read_boo_tmp(domain source) {
+		allow(source, boo.tmp, file, [read]);
```
dburgener, line 60: Does the "this.tmp" case work as well?

I'd think if this line said "this.tmp" instead of "boo.tmp" that we'd expect "this" to still resolve to jkl, despite the cast, which only makes it call boo's function and pass type validation, but still is "really" jkl under the hood.  Is that an accurate understanding of the intent of the code?  And can we add validation to ensure it's the case?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+        }
+        for call in func.original_body {
+            if let Statement::Call(call) = call {
+                if let Some(inner_func) = tmp_functions.get(&call.get_cil_name()) {
+                    if !inner_func.is_castable {
+                        num_changed += 1;
+                        func.is_castable = false;
+                        continue 'outer;
+                    }
+                // The call may be a built-in so check the args
```
dburgener, line 497: This comment confuses me.  This is checking the args regardless of whether it's a built-in or not, right?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+            ));
+        }
+        // Unwrap is safe since we need to go through the loop above at least once
+        return Err(CascadeErrors::from(error.unwrap()));
+    }
+
+    Ok(())
+}
+
+pub fn prevalidate_functions(
```
dburgener, line 619: Would functions.rs be a better home for all of this?

src/functions.rs
```diff
@@ -1304,11 +1304,12 @@ pub struct FunctionInfo<'a> {
     pub is_virtual: bool,
     pub args: Vec<FunctionArgument<'a>>,
     pub annotations: BTreeSet<AnnotationInfo>,
-    original_body: &'a [Statement],
+    pub original_body: &'a [Statement],
     pub body: Option<BTreeSet<ValidatedStatement<'a>>>,
     pub declaration_file: &'a SimpleFile<String, String>,
     pub is_associated_call: bool,
     pub is_derived: bool,
+    pub is_castable: bool,
```
dburgener, line 1312: Does it compile if we make this private?

Either way, can we add a comment that says that this should typically be initialized to true, and prevalidate_functions() will set it to false if needed?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+        }
+    }
+}
+
+// Go through all of the functions and check if they are castable
+// base on functions they call.
+pub fn determine_castable(functions: &mut FunctionMap, types: &TypeMap) -> u64 {
+    let mut num_changed: u64 = 0;
+    // We need tmp_functions to avoid a immutable borrow after a mutable one.
+    let tmp_functions = functions.clone();
```
dburgener, line 483: This is probably a follow-up task since we want to get this working functionally, but this sounds really non-performant.  Since determine_castable() is called in a loop and functions is reasonably sizable, this is a ton of copying.

A future optimization could save a list of names to mark uncastable and then mark them as a cleanup task so the whole below step is done immutably.

I recently filed an issue in our internal issue tracker to add performance benchmarking.  That would be helpful if we go back to optimize so that we can see the impact of a proposed performance optimization like that.

src/compile.rs
dburgener: Okay, looking more closely, I think it's not checking the args on non-built-ins, right?  Since this `else` matches the `if` that's a lookup in the function map.  So if we find it in the function map, we just check whether it's marked as castable or not.  If we don't find it in the function map, we check the arguments.

I see in call_casting.cas a line that says `asd<boo>.some_function(foo); // TODO this should break?` which I think may be exactly this case.  From the comment should I infer it's not breaking now?

I think we need to check the args regardless of whether it's a built-in, because otherwise we're not checking what we pass to non-builtins.

dburgener:
Can you split the commits so that the recursion testing is a separate commit from the casting?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+                        is_term = false;
+                        break;
+                    }
+                }
+            }
+        }
+        if is_term {
+            term_ret_vec.push(func.get_cil_name().clone());
+        } else {
+            nonterm_ret_vec.push(func.clone());
```
dburgener, line 548: Why are we saving the full FunctionInfo on this one instead of just the name?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+    functions: &'a FunctionMap<'a>,
+) -> (Vec<String>, Vec<FunctionInfo<'a>>) {
+    let mut term_ret_vec: Vec<String> = Vec::new();
+    let mut nonterm_ret_vec: Vec<FunctionInfo> = Vec::new();
+
+    for func in functions.values() {
+        let mut is_term = true;
+
+        for call in func.original_body {
+            if let Statement::Call(call) = call {
```
dburgener, line 533: What about calls in other kinds of statements (optional blocks, if blocks etc)?

Something like this is non-terminating:

```
fn foo(...) {
    if (true) {
        something.some_call(...);
    }
}
```

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+    terminated_list: &mut Vec<String>,
+    functions: &mut Vec<FunctionInfo>,
+) -> Result<(), CascadeErrors> {
+    let mut removed: u64 = 1;
+    while removed > 0 {
+        removed = 0;
+        for func in functions.clone().iter_mut() {
+            let mut is_term = true;
+            for call in func.original_body {
+                if let Statement::Call(call) = call {
```
dburgener, line 565: Same thing about if and optional blocks as in initialize_terminated.

Both of these cases should probably be full matches without a `_` case so that we get compiler errors if we add another value to the Statement enum and know to come back and think about the implications here.

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+                    let mut call_cil_name = call.get_cil_name();
+                    // If we are calling something with this it must be in the same class
+                    // so hand place the class name
+                    if call_cil_name.contains("this-") {
+                        if let Some(class) = func.class {
+                            call_cil_name =
+                                call_cil_name.replace("this-", &(class.name.to_string() + "-"))
+                        } else {
+                            return Err(CascadeErrors::from(ErrorItem::make_compile_or_internal_error(
```
dburgener, line 574: I'm trying to think through where this situation arises...  This would happen if we call `this.foo(...)` in a context where `this` isn't a thing (ie at the global scope?)?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+                                "Could not determine class for 'this.' function call",
+                                Some(func.declaration_file),
+                                call.get_name_range(),
+                                "Perhaps you meant to place the function in a resource or domain?")));
+                        }
+                    }
+
+                    if terminated_list.contains(&call_cil_name) {
```
dburgener, line 582: Why not reverse the logic and set is_term=false break if terminated_list does *not* contain the call?  Then we'll short circuit and not keep checking the remaining calls after the first non-terminated one?

src/compile.rs
```diff
@@ -459,6 +459,197 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+    while removed > 0 {
+        removed = 0;
+        for func in functions.clone().iter_mut() {
+            let mut is_term = true;
+            for call in func.original_body {
+                if let Statement::Call(call) = call {
+                    let mut call_cil_name = call.get_cil_name();
+                    // If we are calling something with this it must be in the same class
+                    // so hand place the class name
+                    if call_cil_name.contains("this-") {
```
dburgener, line 569: Rather than doing all the string manipulation manually, I'd just copy this bit from `ValidatedCall::new()`:

```
let tmp_str = match types
                    .get(convert_class_name_if_this(class_name, parent_type)?.as_ref())
                {
                    Some(type_name) => get_cil_name(Some(&type_name.name), &call.name),
                    None => call.get_cil_name(), // Expected to error out below
                };
```

data/policies/call_casting.cas
matt-sheets: That is the case, and I was able to validate.

matt-sheets: It looks like the `this.tmp` is resolved when the macro gets written out to cil.  So by adding:
`jkl<boo>.read_boo_this_tmp(this);` &
```
fn read_boo_this_tmp(domain source) {
    allow(source, this.tmp, file, [read]);
}
```

I get the following cil lines:
`(macro boo-read_boo_this_tmp ((type this) (type source)) (allow source boo-tmp (file (read))))` & `(call boo-read_boo_this_tmp (jkl jkl))`

src/compile.rs
matt-sheets: I made the change to check args all the time, but I do want to talk a bit more about the `asd<boo>.some_function(foo);` call.  I think it should work the way it is written. There are no associated types along that chain.

src/compile.rs
matt-sheets: I put the prevalidate_functions method here since that's where validate_functions was.  Would you want me to move all of those functions(validate_functions, prevalidate_functions, the new functions prevalidate calls) or just the stuff that prevalidate_functions calls?

src/functions.rs
matt-sheets: Private breaks things, but I did add the comment.

src/compile.rs
matt-sheets: https://github.com/dburgener/cascade/issues/138

src/compile.rs
matt-sheets: We need the original_body to look through the statements and this saves us from having to go back to the function map.

src/compile.rs
matt-sheets: If I am iterating through the original_body like I am will I see the something.some_call or do I have to recursively go through the IfBlock?

src/compile.rs
matt-sheets: That or a collection, which I also don't think this is a thing.

src/compile.rs
dburgener: You have to recursively go into the blocks.  The Statement in the original_body will just be IfBlock or whatever, and then that block will contain other statements (IIRC, IfBlock doesn't yet, but will eventually.  Optional should already I believe)

dburgener: We should probably add some sort of helper to handle the block traversal logic.  A function that takes a Vec<Statement> and returns an iterator through all the statements descending into the blocks, defined in ast.rs would be really helpful.

src/compile.rs
dburgener: Oh interesting...  I hadn't realized validate_functions() was here as well.  It does make sense for them to live together...

I think I'd really like both of them to ultimately migrate to functions.rs.  Maybe move the existing validate_functions.rs as its own commit as part of this PR.

src/compile.rs
dburgener: Well, this is another occasion where performance testing would help, but I'd guess that the map lookups are more performant than cloning all the FIs.  If that's the only tradeoff, I'd prefer to save the names and look them up in the map again.

src/compile.rs
dburgener: Okay, yeah that makes sense.

I think this is fine for now.  In the future we might consider if there's a good way to centralize the `this` checking.

That said, do we have to do the string manipulation approach here?  Would convert_class_name_if_this() work here?  It looks like it returns an InternalError.  I'm not sure why, that should maybe become a compile error pointing at the `this`?  That seems like a more centralized approach.

src/compile.rs
matt-sheets: I already started doing this a slightly different way before I saw this comment.  I will upload shortly but can change the approach if you think necessary.

src/compile.rs
matt-sheets: Obsoleted by new function.

src/compile.rs
matt-sheets: We don't have easy access to the TypeInfo, though I assume we can get it.  I did update this to use get_cil_name instead and should be returning a compile error now.

data/policies/call_casting.cas
dburgener: Ah, yeah, that makes sense that that's how it's working now.  I don't think that's the behavior we want though.  I think the expectation is that when I run `jkl<boo>.read_boo_tmp(this)`, I get the access on `jkl.tmp` rather than `boo.tmp`.

The problem is how to achieve that.  The overwrite `this` method works for casting `jkl`, but I don't think we have an immediately obvious way to also convert `boo-tmp` into `jkl-tmp` as well.  I can't immediately think of a way to handle this besides generating a new macro that matches the current one except changing `boo-tmp` into `jkl-tmp`, which isn't the most desirable thing in my mind.  Any ideas of solutions here I'm missing?

src/compile.rs
dburgener: Hmm, I think I may have gotten some_function() and read_boo_tmp_more() mixed up?  I haven't reviewed super thoroughly, but on a quick skim, I think I agree with you now that `asd<boo>.some_function(foo);` looks fine.

data/policies/call_casting.cas
matt-sheets: Unless cil deals with `this` in a sort of find replace kind of way, I agree we would need a new macro for every cast of `read_boo_this_tmp`.  It's a bit out of scope for this work, but if you point me to where we resolve the macros, I can try to find a fix.

src/ast.rs
```diff
@@ -455,6 +455,31 @@ impl Statement {
     }
 }
 
+// The function will take a vector of statements and reduce them down
+// to only their function calls.
```
dburgener, line 459: Probably should expand this comment to call out the fact that it's all possible function calls.  If I write something like:

```
if false {
    some_call(...);
}
```

Then this will still include some_call() even though it's known to be unreachable.  That's certainly the behavior we want here, but probably worth noting in the comment for future reference.

src/functions.rs
```diff
@@ -2674,6 +2706,165 @@ impl<'a> From<&'a FunctionArgument<'a>> for ExpectedArgInfo<'a, '_> {
...
+        removed = 0;
+        for func in functions.clone().iter() {
+            let mut is_term = false;
+            if let Some(function_info) = function_map.get(func) {
+                let func_calls = get_all_func_calls(function_info.original_body.to_vec());
+                for call in func_calls {
+                    let mut call_cil_name = call.get_cil_name();
+                    // If we are calling something with this it must be in the same class
+                    // so hand place the class name
+                    if call_cil_name.contains("this-") {
```
dburgener, line 2814: I think `if call.class_name.as_ref() == Some("this")` would be better.

src/compile.rs
dburgener: I don't think the new update matches the suggestion.  Maybe I'm missing something, but it looks like we're breaking and assume we terminate if we find a single terminating function.  The suggestion above is to assume we are terminating, then if we see a single non-terminating function, set is_term = false and break immediately, which doesn't look like what the code does now.

src/functions.rs
```diff
@@ -2674,6 +2706,165 @@ impl<'a> From<&'a FunctionArgument<'a>> for ExpectedArgInfo<'a, '_> {
...
+    let mut term_ret_vec: Vec<String> = Vec::new();
+    let mut nonterm_ret_vec: Vec<String> = Vec::new();
+
+    for func in functions.values() {
+        let mut is_term = true;
+
+        let func_calls = get_all_func_calls(func.original_body.to_vec());
+
+        for call in func_calls {
+            match call.check_builtin() {
```
dburgener, line 2778: I think we can simplify here with just

```
if call.check_builtin().is_none() {
    is_term=false;
}
```

instead of the full match.  The breaks and continues look unnecessary and the Some(_) case is otherwise empty.

src/functions.rs
```diff
@@ -1849,13 +1859,37 @@ fn validate_cast(
...
+                        "The function is not castable or inherited by caller",
+                    ))
+                }
+            } else {
+                Ok(())
+            }
         }
-        (None, _) => Err(err_ret(start_type.get_range())),
-        (_, _) => Ok(()),
+        (Some(_), Some(_), None) => Err(ErrorItem::make_compile_or_internal_error(
```
dburgener, line 1879: We don't want to expose the term "Function Info" to users.  That's an internal code term that policy developers won't understand.

When can this happen?  It looks to me like the current callers either pass Some(_), Some(_) or None, None for the call and the info.  I think it would just be a compiler bug for them to pass us a call, but no info.  We should probably just document that it's a requirement to be consistent there in the comments for this function and return an InternalError in this case.

src/functions.rs
```diff
@@ -1849,13 +1859,37 @@ fn validate_cast(
...
+            "Func Call was given with no Func Info",
+            file,
+            None,
+            "If you are casting a function you must provide both the call and it's info",
+        )),
+        (Some(_), None, Some(_)) => Err(ErrorItem::make_compile_or_internal_error(
```
dburgener, line 1885: Same as above.

dburgener (changes requested):

src/compile.rs
```diff
@@ -475,6 +476,40 @@ pub fn validate_rules(statements: &BTreeSet<ValidatedStatement>) -> Result<(), C
...
+    while num_changed > 0 {
+        num_changed = determine_castable(functions, types);
+    }
+
+    let (mut terminated_functions, mut nonterm_functions) = initialize_terminated(functions);
+
+    if terminated_functions.is_empty() && !nonterm_functions.is_empty() {
+        let mut error: Option<CompileError> = None;
+
+        for func in functions.values() {
```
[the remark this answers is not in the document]
matt-sheets: You are correct about search_for_recursion catching the error so we don't need this here.

The one thing we don't know here or in search_for_recursion is the nature of the recursive loop.  There may be more than one loop.  We just know that there is a list of functions that did not terminate.

src/compile.rs
matt-sheets: Isn't that the case that if we do find a single terminating thing we are ourselves terminating?  If we find a non-terminating thing there still may be other terminating things to find.  Or do I have my logic just 100% backwards on that?

data/policies/call_casting.cas
matt-sheets: https://github.com/dburgener/cascade/issues/144

src/compile.rs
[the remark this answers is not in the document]
matt-sheets: I updated the code to work the way you outlined, and things are working.  But I just want confirmation that our logic is correct.

src/compile.rs
dburgener: I'm pretty sure you're 100% backwards.  If a function calls any non-terminating functions, then it's non-terminating.

```
fn foo {
    allow(...); // terminating
    allow(...); // terminating
    // etc, more terminating functions
    bar(...); // non-terminating
}

fn bar {
    foo(...); non-terminating
}
```

foo/bar are in a recursive loop, so foo is non-terminating, regardless of the presence of other terminating functions.

src/compile.rs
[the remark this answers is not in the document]
matt-sheets: > I updated the code to work the way you outlined, and things are working. But I just want confirmation that our logic is correct.

Was on the wrong comment sorry.  That was for the:
> Isn't that the case that if we do find a single terminating thing we are ourselves terminating? If we find a non-terminating thing there still may be other terminating things to find. Or do I have my logic just 100% backwards on that?

comment instead.

src/compile.rs
dburgener: At a first read the logic as it stands now seems correct to me.

src/compile.rs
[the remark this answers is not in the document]
matt-sheets: And to clarify something here.  We are not overwriting the error. `add_or_create_compile_error` will either create an error if the error passed in is `None` or it will call `.add_additional_message()`.  The output we will currently get looks something like: 
```
error: Recursive Function call found
   ┌─ data/error_policies/functions_recursion.cas:7:5
   │
 7 │     fn other_read(domain source) {
   │        ^^^^^^^^^^ Calls cannot recursively call each other
   │
   ┌─ data/error_policies/functions_recursion.cas:2:5
   │
 2 │     fn read(domain source) {
   │        ^^^^ Calls cannot recursively call each other
   │
   ┌─ data/error_policies/functions_recursion.cas:12:5
   │
12 │     fn third_read(domain source) {
   │        ^^^^^^^^^^ Calls cannot recursively call each other

```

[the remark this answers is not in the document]
dburgener: OH!  This makes so much more sense now.

Okay, this is fine for now then.  In the future, we could probably enhance the error message to be more helpful by explicitly identifying one or more loops, but this seems good enough for our purposes now.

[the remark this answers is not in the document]
dburgener: Based on that, I'll go ahead and give this another full read-over after lunch.  It seems like all the outstanding issues are addressed, so hopefully it'll all look good at that point.

src/functions.rs
```diff
@@ -1887,17 +1913,12 @@ fn validate_inheritance(
     class_info: Option<&TypeInfo>,
     parent_name: &CascadeString,
     file: Option<&SimpleFile<String, String>>,
-) -> Result<(), ErrorItem> {
+) -> Result<bool, ErrorItem> {
```
dburgener, line 1916: What does this bool mean?

dburgener (approved):

src/functions.rs
matt-sheets: It is true if the all the classes exist and class_info does in fact inherit from the parent_name.  It is false if the classes exist and the class_info does not inherit from parent_name.

I needed something like this because previously this would just error out, and that was fine when castable and inheritance were basically the same.  But now we check if there is inheritance and if there isnt then we check if the function is castable before returning an error.

dburgener: Okay, that seems reasonable.  Can you add a comment on validate_inheritance() about the return type, and specifically mention that `true` means that casting is acceptable and `false` means further checking?

dburgener:
Merged, thanks!

PR #132, opened 2023-02-27
Implement collections to group functions

dburgener:
RFC for now.  We're not really set on the "package" keyword for this at all, and it does somewhat conflict with the casc --package flag, but it's the least bad of the options discussed, so I'm going with that to have something to develop against.


dburgener:
In discussion today, we preferred the "collection" keyword.  The code is changed to reflect that and is now ready for review.

matt-sheets (approved):

PR #134, opened 2023-03-01
Allow the use of associated resources in contexts

dburgener:
Writing the test case for this exposed a bug in nested association extension, so that's fixed in the first commit.

### DIFF
--- a/data/policies/associate.cas
+++ b/data/policies/associate.cas
@@ -77,4 +77,8 @@ virtual domain nest_parent {
 	allow(this, nest_resource, file, write);
 }
 
-domain nest_child inherits nest_parent {}
+domain nest_child inherits nest_parent {
+	extend nest_resource {
+		file_context("/some/path/to/file", [file], system_u:object_r:nest_child.nest_resource:s0);
+	}
+}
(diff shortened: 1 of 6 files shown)


data/policies/associate.cas
```diff
@@ -77,4 +77,8 @@ virtual domain nest_parent {
 	allow(this, nest_resource, file, write);
 }
 
-domain nest_child inherits nest_parent {}
+domain nest_child inherits nest_parent {
+	extend nest_resource {
+		file_context("/some/path/to/file", [file], system_u:object_r:nest_child.nest_resource:s0);
```
matt-sheets, line 82: Is it worth while having more types of statements here? Like allows, functions, "this"?

dburgener: Probably, but it feels a little outside the scope of this PR.  I'm really just trying to test the associated resources in file_contexts aspect here.

Working refpolicy3 compilation, it seems like we may have some other issues around extending associated resources, so I would rather make a follow-up PR that tries to address as many of the extending associated resources issues as I can in one go.  What do you think of that approach?

matt-sheets (approved):

data/policies/associate.cas
dburgener: > Working refpolicy3 compilation, it seems like we may have some other issues around extending associated resource

After far too long unsuccessfully trying to repro those extra issues, I realized they were just another manifestation of the issue fixed here, and this PR addresses what I was seeing.  I'll make another PR in a moment that adds the additional testing suggested here plus one more scenario.

PR #135, opened 2023-03-02
Support port ranges

dburgener:
There's still some weirdness here.  Two outstanding issues that aren't addressed by this PR:

1. We ultimately parse ports from strings into u16s, then convert back to strings, then parse them again.  This is obviously redundant and weird.  To fix it, we'd need to teach Contexts and TypeInstances about ports.  Right now they treat ports as just strings, so everything works fine, but we have to convert back to strings in the middle of everything.
2. The second parse assumes that a list of ports can be comma-separated, which makes no sense whatsoever.  It should be unreachable, since a comma-separated list of ports can't parse.  A list of ports should be a list.  That would parse, but isn't handled by call_to_portcon_rules(). I don't think port lists are used in the current refpolicy3 draft, so I'm punting that for now, but it's something we should loop back to.


src/ast.rs
```diff
@@ -636,13 +636,18 @@ impl Annotations {
...
-    range: Option<Range<usize>>,
+    pub low_port_num: u16,
+    pub high_port_num: Option<u16>,
+    pub range: Option<Range<usize>>,
 }
 
 impl Port {
-    pub fn new(port_num: u16, range: Option<Range<usize>>) -> Self {
-        Port { port_num, range }
+    pub fn new(low_port_num: u16, range: Option<Range<usize>>) -> Self {
```
matt-sheets, line 645: This is a bit of a philosophical question, when defining an object like a port do we prefer to use new or Port {...}?  If we prefer new, I think we should probably have another new that takes in low and high.

src/functions.rs
```diff
@@ -618,32 +622,44 @@ impl PortconRule<'_> {
...
+            Ok(Port {
+                low_port_num: first,
+                high_port_num: Some(second),
+                range: None,
+            })
         }
     } else {
-        port.parse::<u16>().map_err(|_| ())
+        Ok(Port {
+            low_port_num: port.parse::<u16>().map_err(|_| ())?,
```
matt-sheets, line 659: We are dumping the error parse gives in validate_port above.  Is there any worthwhile info in that error that may be useful to pass to the user?  I am guessing it may to be too detailed.

src/ast.rs
dburgener: Broadly speaking, we prefer new(), nearly all the time.  I've been generally avoiding most proper encapsulation in the ast though, because the whole idea is for the parser to shove the code in the ast largely as it exists in source, and then the remainder of compilation is free to muck with it however it likes.  It has been my experience that adding proper encapsulation principles in that narrow use case often adds additional boilerplate and redirection without a ton of benefit, because of how transparent the AST types typically are.

That said, this might be a case where a little bit of extra encapsulation will actually be helpful, because of the combining of individual ports and port ranges into one structure, and because Port is reused as an internal structure as well in functions.rs.  That said, rust doesn't support multiple signatures for the same function, so defining an additional 3 argument new() is not possible, they two news() would need different names (eg new_range()).  Having just typed all that up, new_range() is actually a fine name.  I'm going to go ahead and add it.

src/functions.rs
dburgener: It gives messages like "invalid digit found in string", which seems a little to vague to be useful in my opinion.

matt-sheets (approved):

PR #139, opened 2023-03-06
Support implicit this passing when a function has one argument and the caller passes zero arguments

dburgener:


matt-sheets (approved):

data/policies/implicit_this.cas
```diff
@@ -0,0 +1,10 @@
+resource foo {
+	fn read(domain source) {
+		allow(source, this, file, read);
+	}
+}
+
+domain bar {
+	//equivalent to foo.read(this)
+	foo.read();
```
l0kod, line 9: I the `foo.read(this)` syntax is tested somewhere else, right?

l0kod (changes requested):
I would really like Cascade to forbid function with zero argument, either directly in the lalrpop file by adding a new `Comma` declaration, or in the parser. This is required to avoid implicit argument inconsistency when switching a function from zero argument to one. Anyway, like you said, a function with zero argument doesn't make sense for Cascade.

data/policies/implicit_this.cas
dburgener: Yep, lots of places.  Alphabetically first is here: https://github.com/dburgener/cascade/blob/33ddf1a20e517fcdbfc18261d86360538acc5485/data/policies/alias.cas#L14

dburgener:
@l0kod Added another commit which errors on no arg functions, and an associated test.

l0kod (approved):
Great!

PR #141, opened 2023-03-14
Clean up function argument handling (part 1)

dburgener:
I was hoping to get this all cleaned up comprehensively, including expanded test support, but when handling permission sets in function arguments, which I believe is the last major thing to do, I realized that we need to detect which class a list of permissions belongs to during prevalidate_functions() timing, which is being added in https://github.com/dburgener/cascade/pull/128.

So I'm pushing what I have now, which is independently valuable.  The perm list in a function argument case is broken both before and after this PR, but I'll add that in a follow-on PR.  That PR will also expand the test cases on lists in function arguments more.  I've added a fairly simple test now, but will add more in the next PR.


src/functions.rs
```diff
@@ -415,7 +437,7 @@ fn call_to_fc_rules<'a>(
         )?,
         FunctionArgument::new(
             &DeclaredArgument {
-                param_type: CascadeString::from("obj_class"), //TODO: not really
+                param_type: CascadeString::from(constants::CLASS), //TODO: not really
```
matt-sheets, line 440: This TODO is no longer accurate right?

matt-sheets (approved):

PR #142, opened 2023-03-14
[RFC] Add document detailing Cascade's backwards compatibility commitments post 1.0

dburgener:
This is just an RFC rough draft at this point as a starting point for discussion.


dburgener:
Relevant xkcd, which I would like to explicitly link in the next draft: https://xkcd.com/1172/

PR #146, opened 2023-03-21
Correctly handle duplicate inherited annotations

dburgener:
The first commit fixes certain cases to correctly return a compile error.  The second commit deduplicates cases where the user didn't explicitly cause the conflict, but it was created via multiple inheritance.


data/error_policies/duplicate_association.cas
```diff
@@ -0,0 +1,14 @@
+virtual resource tmp {}
```
matt-sheets, line 1: I am not seeing the `error_policy_test!` for this file

src/internal_rep.rs
```diff
@@ -44,6 +44,77 @@ pub enum AnnotationInfo {
...
+            (Derive(_), Derive(_)) => None,
+            // Enumerate the non-equal cases explicitly so that we get non-exhaustive match errors
+            // when updating the enum
+            (MakeList, _) | (Associate(_), _) | (Alias(_), _) | (Derive(_), _) => None,
+        }
+    }
+
+    // Returns an AnnotationInfo with only the portion in self but not other.
+    // All data should exactly break into three sets: a.difference(b), b.difference(a) and
+    // a.intersection(b) (which is equivalent to b.intersection(a)
```
matt-sheets, line 84: This comment should probably be higher up since it contains information about intersection as well.

matt-sheets (approved):

PR #148, opened 2023-03-22
Fix recursive detection and related errors

matt-sheets:
There were a small number of bugs with the recursion detection which has been resolved as well as improvements with the error given to users

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -2920,13 +2920,25 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
 }
 
 // Helper function to substitue class name if needed
-fn cil_name_helper(call: FuncCall, func_info: &FunctionInfo) -> Result<String, CascadeErrors> {
-    let original_cil_name = call.get_cil_name();
+fn cil_name_helper(
+    call: FuncCall,
+    func_info: &FunctionInfo,
+    function_map: &FunctionMap,
+) -> Result<String, CascadeErrors> {
+    let mut original_cil_name = call.get_cil_name();
+
+    // Deal with casting
+    if let Some(cast_class) = call.cast_name {
+        // This should be mostly safe since we have already looked at castables in prevalidate
+        original_cil_name = get_cil_name(Some(&cast_class), &call.name);
+    }
     // If we are calling something with this or None it must be in the same class
     // so hand place the class name
-    if call.class_name.as_ref() == Some(&CascadeString::from("this")) || call.class_name.is_none() {
+    else if call.class_name.as_ref() == Some(&CascadeString::from("this"))
+        || call.class_name.is_none()
+    {
         if let FunctionClass::Type(class) = func_info.class {
-            return Ok(get_cil_name(Some(&class.name), &call.name));
+            original_cil_name = get_cil_name(Some(&class.name), &call.name);
         } else {
             return Err(CascadeErrors::from(
                 ErrorItem::make_compile_or_internal_error(
@@ -2938,6 +2950,11 @@ fn cil_name_helper(call: FuncCall, func_info: &FunctionInfo) -> Result<String, C
             ));
         }
     }
+    // Finally deal with alias
+    if let Some(call_func_info) = function_map.get(&original_cil_name) {
+        original_cil_name = call_func_info.get_cil_name();
+    }
+
     Ok(original_cil_name)
 }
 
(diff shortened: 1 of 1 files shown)


src/functions.rs
```diff
@@ -2969,19 +3005,58 @@ pub fn search_for_recursion(
...
+                            previous_function
+                        ),
+                    ));
+                    previous_function = function_info.name.clone();
+                }
+            }
+        }
+
         // Unwrap is safe since we need to go through the loop above at least once
         return Err(CascadeErrors::from(error.unwrap()));
```
dburgener, line 3061: I haven't looked at this yet, but I tried running this on my locally modified copy of refpolicy3 and it's panicing on this unwrap.

dburgener: Specifically, what I tested is that I merged this branch with the branch from this PR: https://github.com/dburgener/cascade/pull/147 and built Cascade.

Then I ran it against this refpolicy3 branch: https://github.com/dburgener/refpolicy3/tree/edits-for-recursion

The output after that is:

```
$ ../cascade/target/debug/casc policy/*
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/functions.rs:3061:46
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/functions.rs
```diff
@@ -2919,6 +2919,75 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
     (term_ret_vec, nonterm_ret_vec)
 }
 
+// Helper function to substitue class name if needed
+fn cil_name_helper(
+    call: FuncCall,
+    func_info: &FunctionInfo,
+    function_map: &FunctionMap,
+) -> Result<String, CascadeErrors> {
+    let mut original_cil_name = call.get_cil_name();
```
dburgener, line 2928: I think it would be more idiomatic to do something like:

```
let original_cil_name = if // ... The block starting on 2931 goes here
// after 2952
else {
    call.get_cil_name()
}
```

Then you just assign it once instead of overwriting, and it won't actually need to be mutable (in order to make that happen, the second if block needs to add an else, and return both cases, rather than overwriting).

src/functions.rs
```diff
@@ -2919,6 +2919,75 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
     (term_ret_vec, nonterm_ret_vec)
 }
 
+// Helper function to substitue class name if needed
+fn cil_name_helper(
```
dburgener, line 2923: Can this be named something more descriptive than "helper"?  And a clear comment explaining what it does above and beyond a vanilla "get_cil_name()" would be really helpful.

src/functions.rs
```diff
@@ -2919,6 +2919,75 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
...
+            if call.check_builtin().is_some() {
+                continue;
+            }
+            let call_cil_name = cil_name_helper(call, function_info, function_map)?;
+
+            if terminated_list.contains(&call_cil_name) {
+                continue;
+            } else if visited.contains(&call_cil_name) {
+                break;
+            } else if function_map.get(&call_cil_name).is_none() {
```
dburgener, line 2981: This is the thing we discussed last week where the function doesn't exist, but we don't know that yet, because this is before validate_functions(), right?  Assuming that's accurate, this makes sense, and a comment explaining will be very helpful for future readers.

src/functions.rs
```diff
@@ -2969,19 +3026,58 @@ pub fn search_for_recursion(
...
+            loops.push(find_recursion_loop(
+                func,
+                function_map,
+                terminated_list,
+                &mut Vec::new(),
+            )?);
+        }
+
+        // Now we find the smallest loop and tell that to the user
+        let mut smallest: i32 = -1;
```
dburgener, line 3041: This should be simpler and more readable with min_by https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.min_by

src/functions.rs
```diff
@@ -2969,19 +3026,58 @@ pub fn search_for_recursion(
...
+        for func in smallest_loop {
+            if let Some(function_info) = function_map.get(&func) {
+                // This is the start of the loop, so we want a slightly different message
+                if error.is_none() {
+                    error = Some(add_or_create_compile_error(
+                        error,
+                        "Recursive Function call found",
+                        function_info.declaration_file,
+                        function_info.get_declaration_range().unwrap_or_default(),
+                        "This function is the start of the loop.",
```
dburgener, line 3062: I think it would be helpful to alert the user to the fact that while we're reporting a single loop, there may be others.

dburgener (changes requested):

src/functions.rs
```diff
@@ -2919,6 +2919,84 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
...
+// Search through the non terminating functions to find loops
+fn find_recursion_loop(
+    func: &str,
+    function_map: &FunctionMap,
+    terminated_list: &Vec<String>,
+    visited: &mut Vec<String>,
+) -> Result<Vec<String>, CascadeErrors> {
+    visited.push(func.to_string());
+    if let Some(function_info) = function_map.get(func) {
+        let func_calls = get_all_func_calls(function_info.original_body.to_vec());
```
dburgener, line 2977: This doesn't need to happen as part of this PR, but it looks like we could just convert get_all_func_calls() to take a slice instead of a vec and avoid constructing a vec here.

dburgener (approved):

dburgener:
Merged, thanks!

src/functions.rs
dburgener: https://github.com/dburgener/cascade/pull/151

PR #149, opened 2023-03-30
Adding type and function cast documentation

matt-sheets:


doc/TE.md
```diff
@@ -343,6 +343,52 @@ append.  This could be used, for example to generally allow log files to be
 both writable and appendable, while allowing exceptions for certain append-only
 log scenarios.
 
+## Type casting
+
+To help protect against incorrect AV rules there are rules in place that
```
dburgener, line 348: I'm conflicted on this paragraph.  On one hand, this is a really key motivator for the feature.  On the other hand, there's not really something all that special about AV rules.  Type casting works for any function call.

I might lead with something about how Cascade does strict type checking on all function arguments in order to ensure correctness or something like that.  Then bring up the specific AV rules example as the most likely use case here.  Then finally maybe cite an example of when you would want to use this (probably /proc is the simplest case)

doc/TE.md
```diff
@@ -343,6 +343,52 @@ append.  This could be used, for example to generally allow log files to be
...
+restrict sources to types that have been defined as domains and targets to
+types that have been defined as resources.  If these restrictions need to be
+bypassed type casting can be used.
+
+To cast a type as to something it is not, add the type or attribute you wish
+to cast to in a set of angle brackets after using the type.
+
+```
+domain foo {
```
dburgener, line 357: The problem with this example, is that in this specific case we want people to use `self` instead of `this<resource>`.  I think we should try to have our examples point people in the right direction.  So rather than capability or process rules, I think the example should be reading /proc.

dburgener (changes requested):

dburgener (approved):

PR #156, opened 2023-04-03
Resolve "this" and "this.foo" in function calls.

dburgener:


src/functions.rs
```diff
@@ -2933,43 +2916,49 @@ pub fn initialize_terminated<'a>(functions: &'a FunctionMap<'a>) -> (Vec<String>
...
+    let mut true_call_class = context.convert_arg_this(
+        call.cast_name.as_ref().map(|s| s.as_ref()).unwrap_or(
+            call.class_name
+                .as_ref()
+                .map(|s| s.as_ref())
+                .unwrap_or("this"),
+        ),
+    );
+    if &true_call_class == "this" {
+        true_call_class = match context.get_parent_type_name() {
```
matt-sheets, line 2945: Do we need to worry about casting here?  I don't think we do but I want to double check.

dburgener: Didn't we already handle casting above?  Is there some casting scenario in particular that you're concerned about?

matt-sheets: Sorry miss-read line 2937, thought it said class_name again and thats what your double as_ref comment was about.  You are correct cast should be covered.

matt-sheets (approved):

PR #163, opened 2023-04-10
[RFC] Add ability to do lookups on secondary indices in alias maps

dburgener:
Use this to look up only functions from a specific type, rather than all functions

### DIFF
--- a/src/alias_map.rs
+++ b/src/alias_map.rs
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: MIT
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::ops::Range;
 
 use codespan_reporting::files::SimpleFile;
@@ -12,6 +12,10 @@ pub struct AliasMap<T> {
     declarations: BTreeMap<String, T>,
     #[allow(dead_code)]
     aliases: BTreeMap<String, String>,
+    // Secondary_indices allow efficient lookups based on some other key.  This is useful to
+    // efficiently get a subset of the map based on a prepopulated key, such as all functions for a
+    // given type
+    secondary_indices: BTreeMap<String, BTreeSet<String>>,
 }
 
 pub type AliasMapIter<'a, T> = std::collections::btree_map::Iter<'a, String, T>;
@@ -42,6 +46,7 @@ impl<T: Declared> AliasMap<T> {
         AliasMap {
             declarations: BTreeMap::new(),
             aliases: BTreeMap::new(),
+            secondary_indices: BTreeMap::new(),
         }
     }
 
@@ -78,6 +83,17 @@ impl<T: Declared> AliasMap<T> {
             return Err(error.into());
         }
 
+        for index in value.get_secondary_indices() {
+            match self.secondary_indices.get_mut(&index) {
+                Some(val) => {
+                    val.insert(index);
+                }
+                None => {
+                    self.secondary_indices
+                        .insert(index, BTreeSet::from([key.clone()]));
+                }
+            }
+        }
         self.declarations.insert(key, value);
         Ok(())
     }
@@ -90,6 +106,14 @@ impl<T: Declared> AliasMap<T> {
         self.declarations.values_mut()
     }
 
+    pub fn values_by_index(&self, index: String) -> Vec<&T> {
+        if let Some(secondary) = self.secondary_indices.get(&index) {
+            secondary.iter().map(|v| self.get(v)).flatten().collect()
+        } else {
+            Vec::new()
+        }
+    }
+
     pub fn iter(&self) -> AliasMapIter<'_, T> {
         self.declarations.iter()
     }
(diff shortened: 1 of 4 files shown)


dburgener:
Full system compile     time:   [12.336 ms 12.364 ms 12.399 ms]
                        change: [-2.0204% -1.6903% -1.3454%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 138.5s, or reduce sample count to 10.
Stress functions        time:   [1.3814 s 1.3899 s 1.3984 s]
                        change: [-3.9370% -2.9933% -2.0534%] (p = 0.00 < 0.05)
                        Performance has improved.

I was really hoping for a much better performance win on the "Stress functions" benchmark than this.  But it does seem to consistently give the 3%ish seen here, and the code seems obviously better from a performance standpoint to my eyes.

dburgener:
Looks like stable tests failed an nightly passed on PR checking...  Tests had been passing locally.  I'll investigate more.

dburgener:
I ran callgrind against the function stress test, and prevalidate_functions()/search_for_recursion() dominate there.  Perhaps the win from handle_derive() improvements is being lost against the cost of that.  I'll do the following:

1. Work on getting this to pass tests
2. Look at optimizations of search_for_recursion().

Once both of those are done, then we can re-measure this.  If we get a big win out of search_for_recursion(), then this could become a bigger win percentage-wise.

dburgener:
Just pushed an update that I expect to pass everything except clippy.  #164 should be a big part of the second issue.  I'll retest performance on this after #164 merges and see if that isolates the effect of this change better.

I'm intentionally not fixing clippy yet, since I want to double check the performance impact of clippy's suggestion.  I'll fix or silence it before marking this ready for review.

dburgener:
With taking clippy's suggestion, on one system directly on this PR I got this discouraging (and confusing) result:

Full system compile     time:   [18.414 ms 19.960 ms 21.790 ms]
                        change: [+13.742% +23.877% +35.773%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 174.0s, or reduce sample count to 10.
Stress functions        time:   [1.7004 s 1.7519 s 1.8084 s]
                        change: [+0.8390% +4.6945% +9.1125%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

On my other system, which should be quieter with fewer other programs running I applied clippy's suggestion on top of a merge of this and #164, and got this:

Full system compile     time:   [12.349 ms 12.377 ms 12.409 ms]
                        change: [-4.4579% -4.2176% -3.9794%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 72.9s, or reduce sample count to 10.
Stress functions        time:   [723.65 ms 724.11 ms 724.54 ms]
                        change: [-6.0281% -5.9377% -5.8471%] (p = 0.00 < 0.05)
                        Performance has improved.

That's much more promising.  Intuitively, clippy's suggestion looks better.  I wouldn't be surprised if full optimizations makes them equivalent, but a naive, unoptimized approach would suggest two passes through the list on the non-clippy approach vs one pass on the clippy approach.  I suspect taking the clippy patch is the right thing to do, but I may wait until #164 merges, and rebase this and then retest.  Especially since the results as it stands now are underwhelming.

dburgener:
Okay, I think the issue here is that the benchmark is not sufficiently representative of the actual load this is targeting.  The benchmark is heavy on the inheritance and nested derivation, and that's probably dominating the calculation.  The real target here is refpolicy3, which is a more flat derivation with lots of types deriving functions in a flat structure.  In that situation, when testing a change to make derivation implicit on inheritance, I got "time" numbers like this:

```
real    7m16.670s
user    7m15.903s
sys     0m0.760s
```

With this change (including the clippy fix), that goes to:

real    0m38.160s
user    0m37.409s
sys     0m0.750s

Which is of course much more acceptable.  I'm inclined to say this is worth taking on that basis alone.  Perhaps adding a benchmark that targets this situation better would be helpful.

dburgener:
Latest bench run:

Full system compile     time:   [12.215 ms 12.225 ms 12.237 ms]
                        change: [-0.4285% -0.2960% -0.1512%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 73.6s, or reduce sample count to 10.
Stress functions        time:   [725.48 ms 726.13 ms 726.77 ms]
                        change: [-7.5275% -7.4204% -7.3035%] (p = 0.00 < 0.05)
                        Performance has improved.

matt-sheets (approved):

PR #168, opened 2023-04-12
Add basic initial_context() support

dburgener:
There are a few TODOs left, but this should get basic support in the happy path in.


src/functions.rs
```diff
@@ -961,6 +961,75 @@ fn call_to_fsc_rules<'a>(
...
+    context: &BlockContext<'_>,
+    file: &SimpleFile<String, String>,
+) -> Result<Vec<Sid<'a>>, CascadeErrors> {
+    if context.in_function_block() {
+        return Err(CascadeErrors::from(
+            ErrorItem::make_compile_or_internal_error(
+                "initial_context() calls are not valid in functions",
+                Some(file),
+                c.name.get_range(),
+                "Remove this call",
```
matt-sheets, line 977: add "or move to global scope"?

src/functions.rs
```diff
@@ -961,6 +961,75 @@ fn call_to_fsc_rules<'a>(
...
+        .get_name_or_string(context)?;
+    let sid_context = match Context::try_from(context_str.to_string()) {
+        Ok(c) => c,
+        Err(_) => {
+            return Err(CascadeErrors::from(
+                ErrorItem::make_compile_or_internal_error(
+                    "Invalid context",
+                    Some(file),
+                    context_str.get_range(),
+                    "Cannot parse this into a context",
```
matt-sheets, line 1024: I looked around a bit on this one and we do this exact thing with Context::try_from in a few places.  But I feel like we should be passing up some info in the error we catch here that will help the user out a bit more.  Things like invalid user, role, type, mls at the very least.

I don't vote we fix it in this PR but should go on the TODO pile.

src/functions.rs
dburgener: I thought about that.  The problem is that the most idiomatic thing for a user to do would be to put it inside a type definition.  Global scope is fine of course, but I would expect the common case to be:

```
resource kernel_sid {
    initial_context("kernel", this);
}
```

Maybe this rewrite would be more helpful?  "You may want to place this call directly inside a type definition"

src/functions.rs
dburgener: #171

src/compile.rs
```diff
@@ -1070,7 +1086,20 @@ pub fn get_reduced_infos(
     )?
     .inner(&mut warnings);
 
-    validate_rules(&new_policy_rules)?;
+    // TODO: This is a hack to work around the fact that SIDs contain Contexts, which *might*
```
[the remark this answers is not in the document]
dburgener: #172

src/functions.rs
matt-sheets: I like that better.

PR #169, opened 2023-04-12
Add support for named resource transitions

matt-sheets:


src/functions.rs
```diff
@@ -1139,18 +1161,33 @@ fn call_to_resource_transition<'a>(
...
     }
 
     for file_type in file_types {
-        ret.push(ResourcetransRule {
-            default: Cow::Owned(default.clone()),
-            domain: Cow::Owned(domain.clone()),
-            parent: Cow::Owned(parent.clone()),
-            file_type: Cow::Owned(file_type.clone()),
-        });
+        if obj_name.to_string().is_empty() {
```
dburgener, line 1174: Rather than copy/pasting, let's just do something like:

```
let obj_name = if obj_name.is_empty() {
    None
} else {
    Some(obj_name)
}
```

doc/TE.md
```diff
@@ -440,7 +440,22 @@ Runtime resource transitions are achieved through the resource_transition()
 built in function, which has the following prototype:
 
 ```
-fn resource_transition(resource default, source domain, resource parent, [class] obj_classes);
+fn resource_transition(resource default, source domain, resource parent, [class] obj_classes, string "obj_name");
```
dburgener, line 443: `string obj_name=""`

doc/TE.md
```diff
@@ -440,7 +440,22 @@ Runtime resource transitions are achieved through the resource_transition()
...
+```
+
+* `default` is the type you want the resource to transition to.
+* `domain` is the domain which is creating the resource.
+* `parent` is the type which the resource is being created in.  E.g. The type of
+  the directory a file is being created in.
+* `obj_classes` is the object class(es) of the resource being created.
+* `obj_name` is an optional argument which will only evoke the transition if
+  the name of the resource being created matches `obj_name` exactly.  If this
+  argument is present the resource transition is considered "named".
```
dburgener, line 453: "is present, and not the empty string"

dburgener (approved):

dburgener:
Merged, thanks!

PR #174, opened 2023-04-13
Fix associated call arg parsing

matt-sheets:
Previously associated arg types were forced to be the domain type, this doesn't need to be the case as long as the arg type inherits from domain.  The code now reflects that.

Additionally the error output if you are attempting to call an associated call from a domain with incorrect inheritance has been improved.


src/functions.rs
```diff
@@ -2740,6 +2750,22 @@ fn validate_argument<'a>(
...
+                            ),
+                            arg_typeinfo.get_file().as_ref(),
+                            arg_typeinfo.name.get_range(),
+                            &format!(
+                                "An associated call was made for this domain.  This type should inherit {}",
+                                target_argument.param_type.name
+                            ),
+                        ));
+                    }
+                }
```
matt-sheets, line 2768: This is not as good as it should be.  Right now, we are just pointing to the domain definition and saying it needs to inherit X.  But I am struggling with the best way to get information about the associated call/class here.  Any guidance would be appreciated.

data/policies/associate.cas
```diff
@@ -92,6 +99,16 @@ extend nest_child {
...
 
+virtual domain userdomain {}
+
+@associate([user_tmp])
+domain foo_userdomain inherits userdomain {
+	// Creates new resources foo.tmp and foo.var, and implicitly calls
+	// foo.tmp.associated_call_from_tmp(foo) and foo.var.associated_call_from_var(foo)
+	//
+	// foo.tmp inherits tmp
+	// foo.var inherits var
```
matt-sheets, line 110: Copy/Paste: These comments need to be updated.

data/error_policies/bad_inherits_associated_call.cas
```diff
@@ -0,0 +1,17 @@
...
+	}
+}
+
+@associate([user_tmp])
+domain bad_userdomain {
+	// Creates new resources foo.tmp and foo.var, and implicitly calls
+	// foo.tmp.associated_call_from_tmp(foo) and foo.var.associated_call_from_var(foo)
+	//
+	// foo.tmp inherits tmp
+	// foo.var inherits var
```
matt-sheets, line 16: Copy/Paste: These comments need to be updated

src/functions.rs
```diff
@@ -2740,6 +2750,22 @@ fn validate_argument<'a>(
             if arg_typeinfo.is_child_or_actual_type(target_argument.param_type, types) {
                 Ok(TypeInstance::new(&arg, arg_typeinfo, file, context))
             } else {
+                if let Some(file) = file {
+                    if file.name().is_empty() {
```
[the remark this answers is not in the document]
dburgener: Associated calls have an empty string file name?  That's not something I recall knowing, and probably not something that should be relied upon.

I believe FunctionInfo has an is_associated_call boolean field, which is the canonical way to know if something is an associated call.

I see that we don't have the FunctionInfo in validate_argument().  I suppose we *could* pass it in, but is validate_argument() even the right place to be checking?  Associated calls must have exactly one argument, so if we're validating the Nth argument for N > 1 then we don't need to do this check at all.  Maybe just pulling it up to the validate_arguments() level would make more sense?  Even then we don't have a FI.  Maybe even all the way up to ValidatedCall:new() actually.  We have an FI there, and the other callers of validate_arguments() are builtins, which can't be associated_calls so don't need this check.  That's not deeply thought through, so let me know if I'm missing something.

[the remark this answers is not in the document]
dburgener: Oh, I'm starting to track the logic better.  This is here right now, because we're in the case where validate_argument() has failed.  But since the associated_call was implicit, our generic error message is less than helpful.  So we need to put in a better error message.  The problem is that the FunctionInfo has all the information we need, and we don't have that here.

So the logic about moving it up to ValidatedCall as I outlined above would probably work, but seems redundant and a little weird.  This *is* the right place for error messages.  Maybe adding an Option<FunctionInfo> argument to validate_argument()/validate_arguments() would make sense, and then just passing None everywhere except ValidatedCall::new()?  I'm not wild about continuing to grow the argument list on those already bloated functions, but having the FunctionInfo we're calling seems like a reasonable thing to have while parsing arguments.

src/functions.rs
dburgener: I think most of the issue here is related to the discussion on the other comment.  A lot of what you need will be in the FunctionInfo.  And we should know the domain with the call from the context, I imagine, right?  Between the context and FunctionInfo, I'd think we should end up with enough data to write a helpful error message, once we have the FunctionInfo available at the message spot.

matt-sheets:
New error will look like this:
```
error: Expected type inheriting userdomain for associated call
   ┌─ data/error_policies/bad_inherits_associated_call.cas:11:8
   │
11 │ domain bad_userdomain {
   │        ^^^^^^^^^^^^^^ An associated call: 'associated_call_from_user_tmp' was made for this domain.  That call requires this domain inherit userdomain
   │
   ┌─ data/error_policies/bad_inherits_associated_call.cas:3:18
   │
 3 │ virtual resource user_tmp {
   │                  ^^^^^^^^ Associated class found here

```

dburgener (changes requested):

src/functions.rs
```diff
@@ -2827,6 +2894,64 @@ fn validate_argument<'a>(
...
+                                    {
+                                        let associate_type_info = types.get(resource.as_ref());
+                                        // If for some reason we cannot find the type, its file, or its range, just return the error
+                                        // as is and hope the user can figure it out.  Realistically this should never be the case.
+                                        if let Some(associate_type_info) = associate_type_info {
+                                            if let (Some(file), Some(range)) = (
+                                                associate_type_info.get_file(),
+                                                associate_type_info.get_name_range(),
+                                            ) {
+                                                if let ErrorItem::Compile(e) = error {
```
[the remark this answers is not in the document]
matt-sheets: Unless you can think of another way, I don't think so unless we also pass in the function map.  We have the name of the function, but none of the file/range info.  I thought printing the name, and pointing to the class is a good middle ground.  But if the call comes from an extension of the class it may be harder to find.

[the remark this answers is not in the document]
dburgener: Yeah, those seem like pretty reasonable obstacles.  We can leave this as is for now, and maybe someday that can be a future enhancement.

src/functions.rs
```diff
@@ -2827,23 +2895,99 @@ fn validate_argument<'a>(
...
+// Helper function for handling the error case for validate_argument.
+fn validate_argument_error_handler(
+    arg: ArgForValidation,
+    arg_typeinfo: &TypeInfo,
+    target_argument: &FunctionArgument,
+    types: &TypeMap,
+    file: Option<&SimpleFile<String, String>>,
+    func_info: Option<&FunctionInfo>,
+) -> ErrorItem {
+    // If there is no function call or the call is not an asociated one fall through and return
```
dburgener, line 2920: I don't think "no function call" accurately describes the situation.  And it's mildly confusing, since why are we validating no function.

The situation is that built-ins don't have FunctionInfos, so if func_info is None, that means we're validating a built-in, which can't be an associated call.

Please reword the comment to clarify that None = builtin, which implies not an associated_call.

src/functions.rs
```diff
@@ -2827,23 +2895,99 @@ fn validate_argument<'a>(
...
+            &format!(
+                "An associated call: '{}' was made for this domain.  That call requires this domain inherit {}",
+                func_info.unwrap().name, target_argument.param_type.name
+            ),
+        );
+        // We don't know which class the associated call came from, but that would
+        // be nice information for the user.  So we search through all the associate class
+        // annotations, and find the class we are looking for. Then add that to the
+        // error message.
+        for annotation in arg_typeinfo.get_annotations() {
```
dburgener, line 2942: Can we not point at the associated_call itself?  We're validating it right now, so we should have everything we need, right?  Would that make for a clearer error message?

matt-sheets: If you are talking about pulling it from function info the, it has a declaration file of: `SimpleFile { name: "", source: "", line_starts: [0] }`

dburgener: Oh, right.  If the particular @associated call in this instance is derived, it wouldn't have a valid file.  What this PR does is find the associated resource, but get the name of its parent.

Does this work with a nested syntax like below?

```
domain dom {
    resource res {
        @associated_call
        fn call(userdomain source) {
            allow(source, this, file, read);
        }
    }
}
```

matt-sheets:
```
domain dom {
    resource res {
        @associated_call
        fn call(userdomain source) {
            allow(source, this, file, read);
        }
    }
}
```
When added to assocaite.cas gives the error:
```
error: Expected type inheriting userdomain for associated call
    ┌─ data/policies/associate.cas:118:8
    │
118 │ domain dom {
    │        ^^^ An associated call: 'call' was made for this domain.  That call requires this domain inherit userdomain
```

```
domain dom inherits userdomain {
    resource res {
        @associated_call
        fn call(userdomain source) {
            allow(source, this, file, read);
        }
    }
}
```
Compiles cleanly.

dburgener: Okay, I've been staring at this bewildered for a while, and I think I figured out what's going on.  I think it's because of this line:

https://github.com/dburgener/cascade/blob/5daa36d1e4c455f4108fcc634f1c3ff175c36f65/src/compile.rs#L141

When we create a nested resource, we make a fake parent for it, and that parent is created from a clone of the nested TypeDecl.  That fake parent ends up in this case having the name `res`, while the real associated resource is `dom.res`.  So strictly, the declaration location of `res` shouldn't exist and the declaration location of `dom.res` is the nested declaration.  However, due to the way `res` is made from `dom.res`, they end up pointing at the same location.

The problem here, as @pebenito pointed out, is that if there is a collection named `res` in the same policy, we have an expected conflict.  In order to address this, I'm thinking that I'll want to declare `dom.res` directly in extend_type_map, rather than declaring a fake parent to associate.  I believe that that expected future change will break this.

I *think* the most correct thing to do is to check the file and range on the FunctionInfo we're working on now.  If it's valid, we can add a message and return.  If it's not valid, then that means this \@associated_call was derived.  So we need to locate the parents of func_info.class, and check their \@associated_calls, until we find one that both 1. Has a valid file and range, and 2. Has the type signature we're currently complaining about.  That should theoretically always exist, since if this \@associated_call isn't declared in source, it has to have been derived from somewhere.

src/functions.rs
```diff
@@ -2421,6 +2481,8 @@ pub fn validate_arguments<'a>(
     class_perms: &ClassList,
     context: &BlockContext<'a>,
     file: Option<&'a SimpleFile<String, String>>,
```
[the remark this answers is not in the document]
dburgener: Oh, that's wrong too.  The file in the FunctionInfo is the file where the function is declared.  The file here is the file where it's called.  I guess this whole chain is just wrong.

src/functions.rs
matt-sheets: In the newest push the error now looks like:
```
error: Expected type inheriting userdomain for associated call
   ┌─ data/error_policies/bad_inherits_associated_call.cas:11:8
   │
11 │ domain bad_userdomain {
   │        ^^^^^^^^^^^^^^ An associated call: 'associated_call_from_user_tmp' was made for this domain.  That call requires this domain inherit userdomain
   │
   ┌─ data/error_policies/bad_inherits_associated_call.cas:5:5
   │
 5 │     fn associated_call_from_user_tmp(userdomain source) {
   │        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Associated function found here
```

src/functions.rs
```diff
@@ -2827,23 +2895,117 @@ fn validate_argument<'a>(
...
+            func_info
+                .unwrap()
+                .class
+                .get_name()
+                .unwrap_or(&CascadeString::from(""))
+                .as_ref(),
+        ) {
+            for parent in &class.inherits {
+                if let Some(parent) = types.get(parent.as_ref()) {
+                    if let Some(decl) = &parent.decl {
```
dburgener, line 2967: Rather than going back to the decl, can't we pass in the function map and call `values_by_index()` then check each FunctionInfo?

src/functions.rs
```diff
@@ -2827,23 +2911,109 @@ fn validate_argument<'a>(
...
+        // we are called from.
+        if let Some(class) = types.get(
+            func_info
+                .unwrap()
+                .class
+                .get_name()
+                .unwrap_or(&CascadeString::from(""))
+                .as_ref(),
+        ) {
+            for parent in &class.inherits {
```
dburgener, line 2983: Don't we need to recurse here?  What if the call was derived through a grandparent?

matt-sheets: What would the .cas look like for that?

I added some recursion code to try and fix it and I have something like:
```
virtual domain userdomain {}

virtual resource grandpa_user_tmp {
    @associated_call
	fn associated_call_from_user_tmp(userdomain source) {
		allow(source, user_tmp, file, [read]);
	}
}

virtual resource user_tmp inherits grandpa_user_tmp {}

@associate([user_tmp])
domain bad_userdomain {
	// Errors because bad_userdomain does not inherit userdomain, which
    // the associated call above requires
}
```

This compiles.  In this out.cil I don't see any macro `user_tmp-grandpa_user_tmp-associated_call_from_user_tmp`

dburgener (approved):

dburgener:
Merged, thanks!

PR #176, opened 2023-04-14
[RFC] Support empty lists

dburgener:
The main use case for this is that we want to turn on "implicit derive" in a future PR, where all children that don't use an explicit @derive annotation get an automatic "@derive(*,*)" behavior added internally, so that inheritance automatically gives you the functions you expect without extra boilerplate.  This raises the potential for someone to want to override this behavior.  One obvious way that jumps to mind to do that is "@derive([],[])", meaning "derive nothing please". Alternatives could include the addition of an @noderive annotation, which might be cleaner, particularly if we want to avoid empty list support everywhere.  But maybe we want empty lists anyways.


dburgener:
My current leaning after having implemented this is that `@noderive` might be a cleaner way to handle the "don't derive" use case, and empty lists could easily be error-prone.  There's a lot to be said for just not allowing them.  That said, empty lists are common and useful, and I don't imagine we'd have to work too hard to come up with other use cases, particularly if we start adding "foreach" support or something.

@matt-sheets @l0kod any thoughts?

matt-sheets:
> My current leaning after having implemented this is that `@noderive` might be a cleaner way to handle the "don't derive" use case, and empty lists could easily be error-prone. There's a lot to be said for just not allowing them. That said, empty lists are common and useful, and I don't imagine we'd have to work too hard to come up with other use cases, particularly if we start adding "foreach" support or something.
> 
> @matt-sheets @l0kod any thoughts?

For syntax I like the `@noderive` better than the `@derive([],[])` for general readability.

But with that being said supporting an empty list is probably not a bad thing, _if_ we have a viable use case.  We may run into situation where the code assumes the list has content, but we probably shouldn't be making that assumption anyway and fixing it should be straightforward.

dburgener:
> supporting an empty list is probably not a bad thing, if we have a viable use case

I'm not overly convinced we do.  The two use cases illustrated here are 

1. giving an AV rule no perms, causing to to be removed entirely.  I guess that could potentially be desirable?  It probably needs higher level list processing constructs that we don't have yet to actually be useful.  Maybe something like a call that takes multiple arguments, one of which is a list of perms, the empty list could be used to omit the rules referencing that list?  I'm not convinced that wouldn't be really confusing for users though, and I don't have a concrete "this is why you might want that as a policy writer" sort of use case.
2. The @derive case that's the key motivator, but I think I agree with you that @noderive is more clear

> We may run into situation where the code assumes the list has content, but we probably shouldn't be making that assumption anyway and fixing it should be straightforward.

Yes, that seems mostly fair.  That said, our sexp generating code will happily take empty lists and output invalid CIL from them.  Off the top of my head, I'm not sure where CIL takes an empty list.  So we'd have to take extra care that we never actually output anything from a list that's non-empty.  We maybe should do that, but doing it thoroughly and well and having confidence may be a higher effort than I've put into this PR so far.

PR #177, opened 2023-04-14
Support inheritance from non-virtuals

dburgener:
Everything about inheritance should "just work" (knock on wood), except for the rules.  In other words, type validation, association, derivation etc don't care if the parent is virtual.

The rules are another matter.  Inheriting from a virtual is implemented with the virtual being an attribute, so we get all the rules "for free", by associating the child with the parent and granting access on the attribute.

In the case of a non-virtual, the parent is an actual type, so we can't do that.  One possible approach would be to create an attribute for the parent and grant all parent rules on that instead, but that could be counter-intuitive for binary analysis.  In order to try to be more friendly to cross-language compilation and binary analysis, the approach chosen here was to duplicate all the rules associated with the parent to apply to the child as well.


src/functions.rs
```diff
@@ -2342,6 +2439,44 @@ impl ValidatedCall {
...
+                    // 1. Explicit indexing - This is guaranteed not to panic because new_call.args is
+                    //    a clone of self.args, so they are definitely the same length
+                    // 2. We validated the call already.  Does this child validate?  Should be "yes",
+                    //    because if the parent can validate the child can
+                    new_call.args[index] = CilArg::Name(nv_child.to_string());
+                    // There may be more children to resolve, so we call recursively
+                    // This is guaranteed to terminate, because the typemap has been checked for
+                    // inheritance loops already
+                    let mut recursive_rules = new_call.create_non_virtual_child_rules(types);
+                    //let mut recursive_rules = BTreeSet::new(); // For debugging
```
matt-sheets, line 2466: Looks like a debug line stuck around

matt-sheets (approved):

PR #181, opened 2023-04-19
Automatically @derive(*,*) functions on inherited types.

dburgener:


dburgener:
I'm converting this back to draft.  I'm having some issues with its functionality, and also realized this needs better test coverage.  I'll try to figure out what is causing the functional breakage and add an automatic derive specific test.

dburgener:
Fixed a few bugs, described in the commits.  Marking ready for review again.

src/compile.rs
```diff
@@ -633,6 +631,22 @@ fn derive_functions<'a>(
                     }
                 }
             }
+            if let AnnotationInfo::NoDerive = annotation {
+                saw_no_derive = true;
+            }
+        }
+        if saw_derive && saw_no_derive {
+            // TODO: Add error
```
matt-sheets, line 639: Is this a TODO we want to do now?  Is it something more complicated than just saying we saw a `@derive` and a `@noderive` for type X?

src/functions.rs
```diff
@@ -973,7 +969,7 @@ fn call_to_fsc_rules<'a>(
                     context_range: context_str_arg
                         .get_range()
                         .ok_or_else(|| CascadeErrors::from(InternalError::new()))?,
-                    file: file.clone(),
```
matt-sheets, line 976: The internet is not being super helpful, what is the diffrence between clone and cloned?

src/functions.rs
```diff
@@ -3215,28 +3227,36 @@ pub fn search_for_recursion(
...
+                if let (Some(file), Some(range)) = (
+                    function_info.declaration_file,
+                    function_info.get_declaration_range(),
+                ) {
+                    // This is the start of the loop, so we want a slightly different message
+                    if error.is_none() {
+                        error = Some(add_or_create_compile_error(
+                            error,
+                            "Recursive Function call found",
+                            // TODO: these unwrap_or_default()s seem risky to generate bad messages.
```
matt-sheets, line 3239: I think this todo was done.  I don't see an unwrap_or_default anymore.

src/functions.rs
dburgener: Clone would clone the iterator itself, while cloned is a function on the iterator which clones each element as it iterates:

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned

matt-sheets:
This is a little wonky but here is the fix for 180
https://github.com/dburgener/cascade/pull/188/files

I think we need to talk about this at some point, but that PR works.

Here is a patch file to add the changes to this branch if you want to test independently.
[extend.patch](https://github.com/dburgener/cascade/files/11290618/extend.patch)

matt-sheets (approved):

PR #188, opened 2023-04-20
Fix for extended types can missing annotations

matt-sheets:
Extended types were not getting the correct annotations.  They were getting the annotations from the inherited type alone, not any additional annotaions on the extend block.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -1357,6 +1357,19 @@ fn create_synthetic_resource(
     // Creates a synthetic resource declaration.
     let mut dup_res_decl = class.decl.as_ref().ok_or_else(InternalError::new)?.clone();
     let res_name = get_synthetic_resource_name(dom_info, &class.name);
+
+    if let Some(decl) = &dom_info.decl {
+        for expression in &decl.expressions {
+            if let Expression::Decl(Declaration::Type(d)) = expression {
+                if d.is_extension && d.name == class.name {
+                    dup_res_decl
+                        .annotations
+                        .annotations
+                        .append(&mut d.annotations.annotations.clone());
+                }
+            }
+        }
+    }
     dup_res_decl.name = res_name.clone();
     // See TypeDecl::new() in parser.lalrpop for resource inheritance.
     let mut parent_names = if associated_parents.is_empty() {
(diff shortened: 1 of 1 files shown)


matt-sheets:
I deviated from what we talked about before.  The main reason was that by the time we are in insert_extend_annotations the TypeDecl for the extended type still didnt have any annotations associated with it.  This is because it was grabbing the annotations from its parent, not the extend block.  I am not 100% happy with this, it feels a bit cludgy but it does work.

dburgener:
> I deviated from what we talked about before. The main reason was that by the time we are in insert_extend_annotations the TypeDecl for the extended type still didnt have any annotations associated with it. This is because it was grabbing the annotations from its parent, not the extend block. I am not 100% happy with this, it feels a bit cludgy but it does work.

I'm not really following this.  I'm inclined to agree that the approach here feels kind of cludgy.  And I don't think it works with something like:

```
domain dom {
    resource res {}
}

extend dom.res {
    // ...
}
```

Which, I think is something that should be considered valid.

I don't really understand the issue you're describing above.  Isn't the whole point of insert_extend_annotations() that the TypeDecl doesn't have any annotations?  That's why we saved them to insert later.

Maybe it just needs more cleanup in extend_type_map().  Right now we only handle \@associate annotations, which get inserted on the parent.  But we should be able to grab the annotations on the nested type and directly insert them into the Map extend_type_map() returns for the type itself.

matt-sheets:
I made the changes to insert_extend_annotations to have it run twice.  And I added it around [here](https://github.com/dburgener/cascade/blob/main/src/lib.rs#L184) to match where it was previously called.  The expressions passed into the new PolicyFile, which is then passed into extend_type_map is the first time in compile_machine_policies_internal where bar.tmp is present.  But when I looked into all of the information that is passed into extend_type_map I saw no reference to a noderive association.  So I am not exactly sure where to pull it from to save it off.

So, I dug deeper into where the expressions for bar.tmp where created came up with this solution.

I tried to compile
```
domain dom {
    resource res {}
}

extend dom.res {
    // ...
}
```
And got
```
error: Unexpected character "."
   ┌─ data/policies/extend.cas:36:11
   │
36 │ extend dom.res {
   │           ^ Expected "inherits" or "{"

error: Unexpected character "{"
   ┌─ data/policies/extend.cas:36:16
   │
36 │ extend dom.res {
   │                ^ Expected "(" or "." or "<"
```

matt-sheets:
Abandoning

PR #191, opened 2023-04-21
Set rust MSRV to 1.70

dburgener:
Also update CI to test it and ensure future compatibility


matt-sheets (approved):

dburgener:
One interesting thing that's come to my intention here is that if we set MSRV, clippy won't recommend lints using features above our MSRV.  On one hand, that's reasonable, but on the other, we would like to know about the opportunity to take advantage of new features.  Ideally we could still set MSRV, so it's available for people trying to build Cascade, but also get lint suggestions that would change the MSRV.

This item is about that https://github.com/rust-lang/rust-clippy/discussions/10709#discussioncomment-5721670.  So perhaps that sort of feature is coming soon.

PR #192, opened 2023-04-24
Fix issue with annotations and nested extensions

matt-sheets:
Previously any annotations on a nested extension would be ignored.  The code will now properly propagate the annotations to the types created by the extensions.


src/lib.rs
```diff
@@ -191,6 +196,11 @@ fn compile_machine_policies_internal(
             }
```
matt-sheets, line 196: A few lines above `ww.inner(&mut warnings);` (github isn't letting me put a comment there) should we be calling append_set_map again?  It didn't seem to have any impact when I tried it, but there may be some weird corner case I am nothing thinking about.

src/lib.rs
```diff
@@ -134,7 +135,11 @@ fn compile_machine_policies_internal(
...
         }
 
-        compile::insert_extend_annotations(&mut type_map, extend_annotations);
+        compile::insert_extend_annotations(
+            &mut type_map,
+            extend_annotations.clone(),
```
dburgener, line 140: This seems wrong.  Now the early annotations are being drained from a clone of the map, not the map itself.  Can we modify insert_extend_annotations() to take a mutable reference to the map, rather than taking ownership?

dburgener:
Can we add another test case for something like this?

```
domain foo {
    resource bar {}

    @alias(baz)
    extend bar {}
}
```

And make sure the alias applies appropriately

src/lib.rs
matt-sheets: Adding `debug_assert!(ww.inner(&mut warnings).is_empty());` to replace `ww.inner(&mut warnings);` makes the following tests fail:
```
failures:
    tests::associate_test
    tests::characterization_tests
    tests::invalid_call_casting
    tests::valid_call_casting
```

src/lib.rs
matt-sheets: I have tried both a mutable and non-mutable reference here and got into all sorts of trouble.

If I do a non-mutable reference it complains that:
```
   --> src/compile.rs:256:42
    |
256 |                     t.annotations.insert(a);
    |                                   ------ ^ expected enum `internal_rep::AnnotationInfo`, found `&internal_rep::AnnotationInfo`
    |                                   |
    |                                   arguments to this function are incorrect
```
If I dereference `a` with the above I get:
```
    |
256 |                     t.annotations.insert(*a);
    |                                          ^^ move occurs because `*a` has type `internal_rep::AnnotationInfo`, which does not implement the `Copy` trait
```


If I use a mutable reference I get:
```
   --> src/compile.rs:254:22
    |
254 |             for a in annotations {
    |                      ^^^^^^^^^^^ `std::collections::BTreeSet<internal_rep::AnnotationInfo>` is not an iterator
    |
    = help: the trait `Iterator` is not implemented for `std::collections::BTreeSet<internal_rep::AnnotationInfo>`
    = help: the following other types implement trait `IntoIterator`:
              &'a std::collections::BTreeSet<T, A>
              std::collections::BTreeSet<T, A>
    = note: required for `&mut std::collections::BTreeSet<internal_rep::AnnotationInfo>` to implement `Iterator`
    = note: required for `&mut std::collections::BTreeSet<internal_rep::AnnotationInfo>` to implement `IntoIterator`
```

src/lib.rs
dburgener: fail how?  The debug_assert is firing?  Does it tell you what is in the map at that point?

src/lib.rs
dburgener: I guess what we really need here is [drain_filter](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.drain_filter), but that's nightly only, and likely to remain that way for a while.

I guess the immutable reference option is our best bet.  Just use clone() instead of dereference to get it working, and add a comment on insert_extend_annotations() to convert to drain_filter() once it's stabilized.

src/lib.rs
matt-sheets: Yes the debug assert if failing.

For the data/policies/associate.cas test I get the following when I do `println!("{:?}", ww.inner(&mut warnings));`
```
{CascadeString { string: "bar", range: Some(843..846)
    }: {Associate(Associated { resources: {CascadeString { string: "bin", range: Some(822..825)
                }
            }
        })
    }, CascadeString { string: "foo", range: Some(522..525)
    }: {Associate(Associated { resources: {CascadeString { string: "tmp", range: Some(497..500)
                }, CascadeString { string: "var", range: Some(501..504)
                }
            }
        })
    }
}
```

matt-sheets: For data/policies/call_casting.cas:

```
{CascadeString { string: "boo", range: Some(831..834)
    }: {Associate(Associated { resources: {CascadeString { string: "tmp", range: Some(810..813)
                }
            }
        })
    }
}
```

dburgener: That's weird.  I think those should only arise in nested declarations, and I didn't think apply_associate_annotations() created any nested declarations.  It should all be at the top level I think.  Can you investigate more and understand why these are getting inserted in the first place?

matt-sheets: I looked into the call_casting.cas situation, so any types I reference will be from that file. Primarily I will be referencing boo.tmp:
```
@associate([tmp])
domain boo {
```

When we call `compile::apply_associate_annotations` in `compile_machine_policies_internal` domains like boo.tmp are not part of the type map and their decls have not been made.  The `match associate_exprs` block in `apply_associate_annotations` is the thing that creates the declaration for them, and that declaration gets saved into a new `PolicyFile`.  That new `PolicyFile` is then passed into `extend_type_map` where the types are made from the decls.

This looks like it is working as intended.

dburgener: >  That new PolicyFile is then passed into extend_type_map where the types are made from the decls.

But `extend_type_map()` should only add annotations in one of two cases:

1. A declaration is nested
2. An extend block has an annotation on it

Does `apply_associate_annotations()` generate policy with one of those cases?

matt-sheets: Is boo.tmp considered nested?

This will all be referring to call_casting.cas.
When we call `extend_type_map` for the second time, there are decls for boo, boo.tmp, jkl, and jkl.tmp.  When we iterate through boo we call this block:
```
if !map_annotation_info(
                    &p.file,
                    t.name.clone(),
                    t,
                    &mut warnings,
                    &mut errors,
                    &mut ret,
                )
'''
Which adds the following to ret:
```
{CascadeString { string: "boo", range: Some(831..834) }: {Associate(Associated { resources: {CascadeString { string: "tmp", range: Some(810..813) }} })}}
```

When we call `extend_type_map` for the first time we add the following type but the return BTreeMap is empty.
```
CascadeString { string: "bar", range: Some(17..20) }
CascadeString { string: "foo", range: Some(166..169) }
CascadeString { string: "dom", range: Some(325..328) }
CascadeString { string: "abc", range: Some(397..400) }
CascadeString { string: "xyz", range: Some(479..482) }
CascadeString { string: "asd", range: Some(557..560) }
CascadeString { string: "jkl", range: Some(617..620) }
CascadeString { string: "tmp", range: Some(686..689) }
CascadeString { string: "boo", range: Some(831..834) }
```

Comment (matt-sheets) [reply src/lib.rs:196]: And to be explicitly clear, the TypeDecl expressions for jkl.tmp and boo.tmp are created by `apply_associate_annotations`

Comment (matt-sheets) [reply src/lib.rs:196]: First run with how code appears in this review:
```
Full system compile     time:   [11.525 ms 11.620 ms 11.711 ms]
Found 15 outliers among 100 measurements (15.00%)
  1 (1.00%) low severe
  11 (11.00%) low mild
  3 (3.00%) high mild

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 185.8s, or reduce sample count to 10.
Benchmarking Stress functions: Collecting 100 samples in estimated 185.82 s (100 iterations)
Stress functions        time:   [1.7148 s 1.7209 s 1.7272 s]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
```

Second run with calling append_set_map a second time:
```
Full system compile     time:   [11.244 ms 11.395 ms 11.570 ms]
                        change: [-3.4308% -1.9317% -0.2871%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe

Benchmarking Stress functions: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 184.2s, or reduce sample count to 10.
Stress functions        time:   [1.7959 s 1.8266 s 1.8616 s]
                        change: [+4.2855% +6.1458% +7.9653%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) high mild
  10 (10.00%) high severe
```

Comment (dburgener) [reply src/lib.rs:196]: 100ms!?  How on earth is adding these guys to the map regressing us 100ms?

Stress functions should definitely stress this, because it does a lot of deriving associated resources, but 100ms/6% is a wild amount of regression to see there.

That's more performance cost than I want to pay for something that's just future-proofing and not an actual bug.  Can you please do the following:

1. Don't include the extra `append_set_map()` call here.
2. Do include a comment here explaining how we're dropping Late annotations by not doing so, but the current code doesn't set any Late annotations in `apply_associate_annotations()`, so that's okay, and the performance cost seems to be around 100ms on the Stress Functions test, so the performance penalty isn't worth it
3. File a bug summarizing the same points as 2, and noting that the performance hit is surprising and worth more investigation sometime in the future.

Comment (matt-sheets) [reply src/lib.rs:200]: https://github.com/dburgener/cascade/issues/214

File: data/policies/nested_alias.cas [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+	allow(this, resource, file, [read]);
+
+    @alias(zap)
+	extend tmp {}
+}
+
+domain abc {
+    resource xyz {}
+
+    @alias(bob)
```
dburgener, line 20: Can we add rules to this actually referencing zap and bob so that the test proves out using them as well?

dburgener (approved):

dburgener:
Merged, thanks!

PR #200, opened 2023-04-26
Fix this.* casting resolution

matt-sheets:
Previously if you were trying to cast a this.* object casc would report that was not a castable type even though it is.  This commit fixes that. This commit will also improve the errors given when something cannot be cast.


dburgener (approved):

dburgener:
Merged, thanks!

dburgener:
The fact that we need to redo some of the symbol_in_context() work here seems like a major code smell to me.  I filed https://github.com/dburgener/cascade/issues/201 to try to comprehensively organize our symbol resolution post-0.1.

PR #202, opened 2023-05-01
Defer unresolvable calls up the call stack until they are resolvable

dburgener:


src/functions.rs
```diff
@@ -1508,6 +1527,7 @@ pub struct FunctionInfo<'a> {
     // This will be initialized to true and prevalidate_functions will set this
     // to false if needed.
     pub is_castable: bool,
+    pub callers: BTreeSet<CallerInfo>,
```
matt-sheets, line 1530: Would this have a chain of all callers (ancestors) of a given function or just the immediate callers(parents)?

I ask because if it is immediate callers, we may be able to use the caller name to map it to another function info, which may make the recursive check more efficient.

src/functions.rs
```diff
@@ -1902,6 +1924,25 @@ impl<'a> FunctionInfo<'a> {
...
+    // passed into it.  If the symbol is not the name of a function argument, return None
+    pub fn symbol_to_caller_symbol(
+        &self,
+        symbol: &str,
+        caller_args: &Vec<CilArg>,
+    ) -> Option<String> {
+        for (func_arg, caller_arg) in self.args.iter().zip(caller_args) {
+            if func_arg.name == symbol {
+                // We ignore perm lists, for no good reason.  This should probably return a CilArg
+                // and let the parent unpack it
```
matt-sheets, line 1938: Worth opening an issue to track?

src/functions.rs
```diff
@@ -2393,13 +2428,206 @@ impl TryFrom<&ValidatedStatement<'_>> for sexp::Sexp {
...
+pub enum DeferredStatement {
+    Call(DeferredCall),
+    Validation, // TODO
+}
+
+impl DeferredStatement {
+    // Copy into the parent context, converting argument names
+    fn parent_copy(&self, caller_args: &Vec<CilArg>, orig_function: &FunctionInfo) -> Self {
+        match self {
+            // TODO: 0
```
matt-sheets, line 2449: 0?

data/expected_cil/arg_call.cil
```diff
@@ -0,0 +1,183 @@
...
+(typeattributeset resource (bar2))
+(type bar3)
+(roletype object_r bar3)
+(typeattributeset foo (bar3))
+(typeattributeset resource (bar3))
+(macro bar1-read ((type this) (type source)) (allow source this (dir (read))))
+(macro bar2-read ((type this) (type source)) (allow source this (lnk_file (read))))
+(macro bar3-read ((type this) (type source)) (allow source this (chr_file (read))))
+(macro baz-call_source_read ((type this) (type source) (type arg)) 
+;Pushed to callers: (source-read foo arg)
```
matt-sheets, line 169: I really like we leave a bread crumb like this.  Is there any way to put this on the previous line with the macro?  If this is any significant amount of work don't do it; it's not hard enough to parse.

src/functions.rs
```diff
@@ -2393,13 +2428,206 @@ impl TryFrom<&ValidatedStatement<'_>> for sexp::Sexp {
...
+impl From<&DeferredCall> for ValidatedCall {
+    fn from(dc: &DeferredCall) -> Self {
+        ValidatedCall {
+            cil_name: get_cil_name(Some(&dc.call_class_name), &dc.call_func_name),
+            args: dc.args.clone(),
+        }
+    }
+}
+
+// TODO: Clean up unwraps.
```
matt-sheets, line 2584: TODO

src/functions.rs
matt-sheets: From the propagate code below it looks like its just the parents.

If that is the case I think it should be "easy" to get a call stack and just make sure the function we are validating doesn't appear in it.

src/compile.rs
```diff
@@ -614,6 +630,27 @@ pub fn validate_functions<'a>(
     errors.into_result(WithWarnings::new(functions, warnings))
 }
 
+fn postvalidate_functions(
+    functions: &mut FunctionMap,
+    types: &TypeMap,
+) -> Result<(), CascadeErrors> {
```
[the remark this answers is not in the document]
dburgener: I wrote that first and assumed I'd end up with errors to return.  I guess I'll get some if I turn all the `propagate` unwraps into internal errors though.

src/functions.rs
dburgener: Awesome, that would be a good optimization. We'll need to make sure that nothing in validate_functions() relies on the assumption that there's no recursion, since we'd be moving the check from prevalidate to postvalidate.

src/functions.rs
dburgener: You don't see any '0' because I did the TODO :)

data/expected_cil/arg_call.cil
dburgener: It would be pretty straightforward to make it go on the end of the line.  The function would still be split up though because the closing paren is on the next line, and that has to go there so it's not in the comment.

I kind of like putting them on their own line, so that it's easy to see the exceptional bit (the comment)

src/functions.rs
dburgener: Filed https://github.com/dburgener/cascade/issues/208

src/functions.rs
matt-sheets: Looks like you got the unwraps but comment is still here.

dburgener:
FYI, I'm seeing a regression in refpolicy3 compile on this branch, probably introduced by 4460f42404ce0d4f61ab8111833e308b293fe5aa.  Will try to get that sorted out and notify here once it's fixed.

dburgener:
Pushed the fix.  The change is in the commit now hashed as 31786811df34337801d09e57254f871f1c0c0d9b

matt-sheets (approved):

PR #205, opened 2023-05-01
Clean up type validation on lists

dburgener:
https://github.com/dburgener/cascade/pull/199 caused some of our type matching on lists to start to fail.  First clean up the error message when this happens, which was previously a TODO, then try harder to infer the right types.  As mentioned in the second commit message, this has us guess (possibly wrongly) on ambiguity.  For a post-0.1 future enhancement, we should detect ambiguity and require the developer to annotate.

### DIFF
--- a/data/policies/permissions.cas
+++ b/data/policies/permissions.cas
@@ -11,6 +11,13 @@ domain foo {
 	allow(this, self, capability, bpf);
 	allow(this, self, bpf, map_create);
 	allow(this, read, file, read);
+
+	allow(this, read, file, some_perms);
+	allow(this, read, file, some_perms_other_order);
 }
 
 resource read {}
+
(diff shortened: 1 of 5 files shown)


data/policies/permissions.cas
```diff
@@ -14,3 +14,7 @@ domain foo {
 }
 
 resource read {}
+
+let some_perms = [ write read getattr ];
+
+let some_perms_other_order = [ read write setattr ];
```
matt-sheets, line 20: Did you intend for these to be used in this cas file or just make sure they don't break compile?

dburgener: I was just making sure they don't break compile.  It would probably make sense to use them as well though so that the test validates that they were actually processed correctly.

PR #211, opened 2023-05-02
Remove unecessary for loop

dburgener:
After subsequent refactoring, we no longer test the call and cast names separately.  We resolve first, then test.  So instead of looping, resolve, and if it doesn't exist, calculate the range to report on


dburgener:
Thanks for the review.  I expect this to conflict with https://github.com/dburgener/cascade/pull/202, so I'm going to wait to merge this until after the big and important one has merged.  Then I'll fix up the conflict here so this can get merged.

PR #216, opened 2023-05-02
Drop unicode aware regexes

dburgener:
The regex tokens "\s" and "\d" are "unicode friendly" and could match whitespace and digit characters in other encodings than ASCII.  The result of including them in our grammar is that we then require regex support for the entirety of unicode, which builds a large table of unicode characters into our binary.

Replace \s and \d with their ASCII-only equivalents, [[:space:]] and [[:digit:]].

Note that this does not remove our unicode dependency by itself. Lalrpop 0.19.12 builds in unicode support by default because it doesn't yet expose the correct configuration knobs to users to enable or disable unicode support.  Once lalrpop 0.20 is released, then users that don't require unicode support will be able to disable it using a feature. This change moves Cascade to that "users that don't require unicode" group, so that we can take advantage of removing unicode support once Lalrpop exposes the correct knob.

### DIFF
--- a/src/parser.lalrpop
+++ b/src/parser.lalrpop
@@ -330,11 +330,11 @@ IPv6: CascadeString = {
 
 // lexing precedence
 match {
-	r"\s*" => { },
+	r"[[:space:]]*" => { },
 	r"//[^\n\r]*[\n\r]*" => { },
-	r"(\d{1,3}\.){4}" => IPv4Regex,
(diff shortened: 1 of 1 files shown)


dburgener:
Rebased and published this.  Since 0.1 is delayed, I think it makes sense to take this for 0.1.

matt-sheets (approved):

PR #227, opened 2023-05-04
Now return a warning if casting cannot find a type

matt-sheets:


dburgener:
This looks good.  Can you also add a line to data/policies/nonexistent_optional.cas exercising it so we have an explicit positive test?

dburgener (approved):

PR #239, opened 2023-05-23
Fix issues around deriving types with aliases

dburgener:
There were probably a few bugs in here, but basically the issue is around deriving functions with combinations of real functions and aliases sharing the same name.  The new policy in
data/policies/derive.cas is the motivating example.  The alias weirdness in data/policies/automatic_derive.cas is relevant as well, and I think was semi-broken with two bugs canceling each other out before.

The three bug fixes here are:

* Derive only from *direct* parents, otherwise we lose overrides at the child level
* Include the real class name in class aliases, otherwise we drop function aliases when the class has no aliases
* If we're deriving a function with aliases to either something we already have declared explicitly, or that we are also deriving, drop those aliases


src/functions.rs
```diff
@@ -1931,14 +1932,27 @@ impl<'a> FunctionInfo<'a> {
...
             for func_alias in &derived_name_aliases {
                 // No alias for real class and func combo
-                if *class_alias == &deriving_type.name && name == func_alias {
+                if class_alias == &deriving_type.name && name == func_alias {
+                    continue;
+                }
+                // If we already define a function with this name, then we're overriding that
+                // alias, so we don't alias to it here. This includes functions we're in the
+                // process of deriving
+                if deriving_type.defines_function(func_alias.as_ref(), functions)
```
matt-sheets, line 1948: Since both of these ifs lead to a `continue` we could combine them, but I get if you want to separate them for readability.

dburgener: Yeah, I think it's more readable to leave them separate.  It becomes quite the elaborate conditional if we combine, and this way it's natural to keep the comments separate as well for the two separate cases.

PR #240, opened 2023-05-26
Don't require a global type for nested associations

dburgener:
1. Declare nested resources directly in extend_type_map()
2. Use a new NestAssociation internal-only annotation to track
association inheritance for them
3. Call *all* associated calls once we're all done with types.  This
allows catching the nested and annotated all at once
4. Fix a *lot* of bugs around nested declarations


src/compile.rs
```diff
@@ -160,25 +160,38 @@ pub fn extend_type_map(
...
                                 }
                                 Err(e) => errors.append(e),
                             }
+                        } else {
+                            // The domain doesn't exist in the map yet, so we can't use
+                            // make_duplciate_associate_error() here
+                            return Err(CompileError::new(
+                                "This resource is explicitly associated to both the parent and child.  (Perhaps you meant to extend the existing resource in the child?)",
+                                &p.file,
+                                associated_type.name.get_range().unwrap(), // TODO: remove unwrap
```
matt-sheets, line 190: TODO

src/compile.rs
```diff
@@ -1504,17 +1535,37 @@ fn create_synthetic_resource(
         ));
     }
 
+    let class_name = if class.is_associated_resource(types) {
+        // TODO, this can be a lot cleaner if split *is* the check for .
+        class.name.as_ref().split_once('.').unwrap().1.into()
```
matt-sheets, line 1540: If we don't have a `.` than this unwrap would be unsafe correct?  How sure are we that we will have a `.`

src/compile.rs
```diff
@@ -1689,7 +1697,32 @@ fn interpret_associate(
                     global_exprs,
                     extend_annotations,
                 ) {
-                    Ok(_) => {}
+                    Ok(_) => {
+                        // If the associated call is derived, we'll need to add it in later.  If
+                        // the association is inherited, we need to make sure to mark it now so we
+                        // know to do that.
+                        let class_name = if class.is_associated_resource(types) {
+                            class.name.as_ref().split_once('.').unwrap().1.into()
```
matt-sheets, line 1705: Same unwrap comment as above.

src/compile.rs
```diff
@@ -1907,9 +1934,18 @@ fn inherit_annotations<'a>(
         dom_info
             .annotations
             .iter()
-            .map(|a| InheritedAnnotation {
-                annotation: a.clone(),
-                parents: vec![dom_info],
+            .map(|a| {
+                let ann = if let AnnotationInfo::NestAssociate(r) = a {
+                    //let resources = r.resources.iter().map(|r| get_synthetic_resource_name(&dom_info.name, r)).collect();
+                    //AnnotationInfo::Associate(Associated {resources})
```
matt-sheets, line 1940: Delete.

src/compile.rs
```diff
@@ -2084,49 +2118,74 @@ pub fn call_derived_associated_calls<'a>(
...
-                            // These are definitely ValidatedCalls, but ValidatedCall::new()
-                            // returns ValidatedStatements, so they can be inserted directly
-                            for c in validated_calls {
-                                ret.insert(c);
+        for f in funcs.values() {
+            if f.is_associated_call {
+                let resource_name = match f.class {
+                    FunctionClass::Type(n) => n.name.clone(),
+                    _ => {
+                        // Can't derive from Global or API
```
matt-sheets, line 2127: Would this be something user is trying to do erroneously that we should warn?

src/compile.rs
```diff
@@ -2084,49 +2118,74 @@ pub fn call_derived_associated_calls<'a>(
...
+                        Some((_, r)) => r,
+                        None => resource_name.as_ref(),
+                    };
+                    let call = make_associated_call(
+                        get_synthetic_resource_name(&t.name, &r_basename.into()),
+                        f,
+                    );
+                    let args = vec![FunctionArgument::new_this_argument(t)];
+                    // TODO: Should there be a parent_context here?  I think this is
+                    // effectively a "fake" context since we're not really parsing the tree
```
matt-sheets, line 2166: What are you thinking the parent context should be?

src/compile.rs
dburgener: `is_associated_resource()` does `.contains('.')`.  That says, I added the `basename` function that should be a lot cleaner here, rather than checking and splitting.  I'll update to that.

src/compile.rs
dburgener: So the only real change here is indent.  The comment existed both with and without this change.

I'm not really sure that having a real context actually makes sense though.  Having a context for the domain would let us handle let bindings from the domain.  But we may not want that anyways.  I'm not sure that something other than this fake context is really needed?  Idk...

src/compile.rs
dburgener: As with below, the change here is just indent, so the comment is old.

I don't think it's possible to erroneously try to derive from a global function.  You have to inherit the parent to derive the function.  In terms of collections (not API anymore), I'd imagine there is/should be checks elsewhere.  Seems outside the scope of this PR though.  I'd think we'd warn on those back when we're parsing annotations on functions.

matt-sheets (approved):

PR #250, opened 2023-09-26
Fix issues with nest association conflicting with regular association

dburgener:


src/compile.rs
```diff
@@ -1592,40 +1597,28 @@ fn create_synthetic_resource(
...
-        });
-    if !global_exprs.insert(Expression::Decl(Declaration::Type(Box::new(
-        dup_res_decl.clone(),
-    )))) {
+    new_decl.annotations.annotations = classes
+        .iter()
+        .flat_map(|c| c.decl.iter().flat_map(|d| d.annotations.annotations.iter()))
+        .filter(|a| a.name.as_ref() == "makelist" || a.name.as_ref() == "derive")
+        .cloned()
+        .collect(); // TODO: dedup?
```
matt-sheets, line 1615: Looks like we don't need the de-dup?

src/compile.rs
```diff
@@ -1695,52 +1688,87 @@ fn interpret_associate(
...
+        // be `ancestor.resource`, and we don't know which, or which ancestor.  And multiple
+        // inheritance is possible.
+        // Okay, so we use the associated resources as planned.  Basically exactly the original
+        // logic.  The objection to the objection is that the parents aren't necessarily the ones
+        // that have the real resources
+        // What the above is missing is that for dom.res, res may or may not exist.  Actually,
+        // existance is semi-irrelevant.  We need to know if we associated the one that exists,
+        // specifically.  Which, I guess means that real_parent_resources maybe also need to be
+        // parents?  The alternative, which may be cleaner is just to save in AssociatedResource
+        // whether we have the base or not.
```
matt-sheets, line 1719: This seems like a "talk thing through" comment.  I would clean up.

src/compile.rs
```diff
@@ -1695,52 +1688,87 @@ fn interpret_associate(
...
+
+        let mut classes = Vec::new();
+        for real_parent_resource in res.get_class_names() {
+            match types.get(&real_parent_resource) {
+                Some(class) => classes.push(class),
+                None => {
+                    // TODO: This is kind of a weird error in the nested case, since the full name
+                    // doesn't exist in source
+                    // I think AssociatedResource needs to be augmented to handle having the
```
matt-sheets, line 1728: Candidate for issue?

src/compile.rs
```diff
@@ -2140,16 +2168,27 @@ pub fn call_associated_calls<'a>(
...
                         AnnotationInfo::NestAssociate(associations) => {
-                            if associations.resources.iter().any(|r| r == &resource_name) {
+                            // This needs a function in AssociatedResource.
+                            // Is there a gotcha to implementing ParitalEq (and Eq???) for
+                            // AssociatedResource and CascadeString?
+                            // Not Eq I think, because Eq promises standard equality properties
+                            // like transitivity, and this relation wouldn't be transitive
+                            // Honestly, implementing PartialEq, when the relation is decidedly not
+                            // Eq feels a little icky.  This should probably just be a normal
+                            // member function
```
matt-sheets, line 2186: This seems like a "talk thing through" comment. I would clean up.

src/internal_rep.rs
```diff
@@ -500,16 +618,52 @@ impl TypeInfo {
...
+            | AnnotationInfo::NestAssociate(associations) = ann
+            {
                 for res in &associations.resources {
-                    if res.as_ref() == associate_name && res.get_range().is_some() {
-                        return res.get_range();
+                    if res.string_is_instance(&CascadeString::from(associate_name))
+                        && res.get_range(associate_name).is_some()
+                    {
+                        // I think the issue is that this get_range() is returning None
+                        // In other news, I'm an idiot.  We are in a block for an if .is_some() on
```
matt-sheets, line 632: Is this still the case with the `get_range(associate_name)` call?

src/internal_rep.rs
```diff
@@ -500,16 +618,52 @@ impl TypeInfo {
...
+                ar.get_range(
+                    get_synthetic_resource_name(&self.name, &associate_name.into()).as_ref(),
+                )
+            });
+
+        if ar_range.is_some() {
+            return ar_range;
+        }
+
+        // If the resource is foo.bar, and we are foo, we need to check bar as well.  If the reason
```
matt-sheets, line 653: Did you mean resource instead of reason at the end of this line?

matt-sheets (commented):
Initial review. I will try to go over it again to understand the logic better.

src/compile.rs
```diff
@@ -1528,57 +1528,62 @@ fn generate_type_no_parent_errors(missed_types: Vec<&TypeInfo>, types: &TypeMap)
...
+        // The key thing here is the ranges of each class, which the associated resource struct
+        // owns.  Perhaps just passing it in wholesale makes sense?
         return Err(ErrorItem::make_compile_or_internal_error(
             "not a resource",
             dom_info.declaration_file.as_ref(),
-            class_string.get_range(),
+            classes
+                .iter()
+                .find(|c| !c.is_resource(types))
+                .and_then(|c| c.name.get_range()),
```
matt-sheets, line 1562: I _think_ this will just return the first class that is a not a resource in the Vec.  I think you should be able to use maybe_add_additional_message to get all of them in one pass.

src/compile.rs
dburgener: Well, at a minimum we pass the existing tests without deduping.  The test coverage with annotations is pretty poor though.  I'm thinking that an issue for later follow-up to both extend test coverage and look into this makes sense.

I do think that duplicate `@makelist`s are just wasting storage.  I suppose duplicate `@derive`s might cause it to derive the same function twice, which would be a problem.

src/compile.rs
dburgener: Hmm, the error actually looks fine to me now...  I probably wrote that comment earlier in development and should have deleted it.  The second two lines about augmenting AssociatedResource is definitely done.

Probably what I should do here is add another error test case to exercise this code path and make sure it looks right, then delete the comment.

src/internal_rep.rs
dburgener: Oh, I definitely intended to remove the "I'm an idiot" comment :)

This (the comment) was all just wrong from the get-go, as explained in the comment.  I'll delete.

src/compile.rs
dburgener: I was going to do what I said above, but I actually don't think this path is reachable in the nested case.  The nested method of association actually creates the resource, so you can't have it not exist.  And even if it were reachable, it should be fine with the AssociatedResource augmentation that is done.  So I'm going to just delete this comment.

matt-sheets (approved):

PR #259, opened 2023-10-06
Resolve the type field in full contexts

dburgener:


src/internal_rep.rs
```diff
@@ -1100,6 +1100,26 @@ impl<'a> Context<'a> {
...
+                    if self.setype == "this" {
+                        block_context
+                            .get_parent_type_name()
+                            .map(|cs| Cow::Owned(cs.to_string()))
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or_else(|| self.setype.clone()),
+            ..self.clone()
```
matt-sheets, line 1120: What is the double '.' here?

matt-sheets (approved):

src/internal_rep.rs
dburgener: It's the "struct update syntax".  https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

Basically, it just means "clone the equivalent field from `self` for the remaining fields`

PR #278, opened 2025-02-05
Test against selinux userspace 3.8

dburgener:
This was recently released


.github/workflows/restore-test.sh
```diff
@@ -1,13 +1,31 @@
 #!/bin/bash
 
-set -e -u -o pipefail
+set -e -u -x -o pipefail
```
matt-sheets, line 3: Did you want to leave the -x in?  Same comment for test.sh

matt-sheets (approved):

.github/workflows/restore-test.sh
dburgener: Yeah, I was a little surprised they weren't in already, because it's always helpful for debugging CI failures.  I feel like the cost is low and the value is high for having the additional debug insight.

